PR #14, opened 2018-01-13
Errors...

### LINKED ISSUE #4: Warn if execution time is very fast
... it could be affected/limited by the intermediate shell

hmenke:
- if execution is faster than 1 ms
- if command is empty

Fixes #4

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -4,12 +4,20 @@ extern crate clap;
 extern crate indicatif;
 
 use std::cmp;
+use std::error::Error;
+use std::io;
 use std::process::{Command, Stdio};
 use std::time::Instant;
 
 use indicatif::{ProgressBar, ProgressStyle};
 use ansi_term::Colour::{Cyan, Green, Red};
 use clap::{App, AppSettings, Arg};
+
+/// Print error message to stderr and terminate
+pub fn error(message: &str) -> ! {
+    eprintln!("{}", message);
+    std::process::exit(1);
+}
 
 struct CmdResult {
     /// Execution time in seconds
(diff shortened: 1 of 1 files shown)


sharkdp:
Welcome, and thank you very much!

I'm going to merge this as is, but I will likely change a few things immediately because this project is in a really early stage.. I'm surprised there are already some visitors :+1: :smile: 

> Errors ... if execution is faster than 1 ms

I will probably turn this into a warning that would be displayed after the statistics, such that subsequent commands would still be benchmarked.

> Errors ... if command is empty

Thanks, good catch!

hmenke:
You're welcome.

Actually the program runs fine even with an empty command, i.e. `hyperfine ""`, though everything that it is doing at that point is benchmarking the shell.

sharkdp:
> That's because I follow you...

:open_mouth: :smile: 

By the way... greetings! You are now in New Zealand?

> Actually the program runs fine even with an empty command, i.e. hyperfine "", though everything that it is doing at that point is benchmarking the shell.

that makes me wonder...

... we could actually use this to gauge the time measurements by running a benchmark on the shell-spawning itself and a subsequent subtraction of that time from the program execution times.

On the other hand, this would only help for really fast execution times and they are problematic for various reasons. Let's go with the warning for now..

hmenke:
> You are now in New Zealand?

Indeed.

> we could actually use this to gauge the time measurements

Then it would probably be easier to cut out the intermediate shell altogether.  Why is it even necessary?  Do you want to benchmark shell scripts?  In this case you could do `hyperfine "sh -c 'echo Hello World'"` but I guess this is not such a common use-case.

sharkdp:
> Then it would probably be easier to cut out the intermediate shell altogether. Why is it even necessary?

Mainly because it improves UX (in my opinion - because you have all the power of the shell without having to add `sh -c "..."`) and makes the command line parsing easier.

If we would cut out the intermediate shell, we would have to either:

* "Parse" (a subset of) shell syntax to split up the commands like `find -name "pattern with spaces"` into multiple arguments (`find`, `-iname`, `"pattern with spaces"`) for `Command::new`. You might also want to substitute common arguments like `~`.
* Invent a `find`-like syntax.. `hyperfine --command1 program1 -a -b --command2 program2 -c -d`

PR #25, opened 2018-01-22
Recommend use of `sudo -v` over `sudo echo`

half-cambodian-hacker-man:
`sudo -v` (or `sudo --validate`) updates the sudo credentials cache without running any commands.

It's slightly safer than `sudo echo` under the unlikely chance that `~/.local/bin/echo` (or elsewhere on the PATH) could be set to something malicious.


sharkdp:
Thank you very much!

I did not know that.

PR #29, opened 2018-01-30
added void linux install to readme

wpbirney:
added hyperfine to the void linux repo, can now be installed via void's xbps

thanks for your awesome work on hyperfine!


sharkdp:
Fantastic, thank you very much!

Issue #28, opened 2018-01-30
Unresolved import `libc::rusage`


TheInitializer:
```
λ cargo install hyperfine
    Updating registry `https://github.com/rust-lang/crates.io-index`
 Downloading hyperfine v0.4.0
  Installing hyperfine v0.4.0
 Downloading colored v1.6.0
 Downloading [...]
 Downloading rustc-serialize v0.3.24
   Compiling strsim v0.6.0
   Compiling [...]
   Compiling hyperfine v0.4.0
error[E0432]: unresolved import `libc::getrusage`
 --> .cargo\registry\src\github.com-1ecc6299db9ec823\hyperfine-0.4.0\src\hyperfine\cputime.rs:1:12
  |
1 | use libc::{getrusage, rusage, RUSAGE_CHILDREN};
  |            ^^^^^^^^^ no `getrusage` in the root

error[E0432]: unresolved import `libc::rusage`
 --> .cargo\registry\src\github.com-1ecc6299db9ec823\hyperfine-0.4.0\src\hyperfine\cputime.rs:1:23
  |
1 | use libc::{getrusage, rusage, RUSAGE_CHILDREN};
  |                       ^^^^^^ no `rusage` in the root

error[E0432]: unresolved import `libc::RUSAGE_CHILDREN`
 --> .cargo\registry\src\github.com-1ecc6299db9ec823\hyperfine-0.4.0\src\hyperfine\cputime.rs:1:31
  |
1 | use libc::{getrusage, rusage, RUSAGE_CHILDREN};
  |                               ^^^^^^^^^^^^^^^ no `RUSAGE_CHILDREN` in the root

error: aborting due to 3 previous errors

error: failed to compile `hyperfine v0.4.0`, intermediate artifacts can be found at `C:\Users\dkter\AppData\Local\Temp\cargo-install.2sl6wvAhi3wL`

Caused by:
  Could not compile `hyperfine`.

To learn more, run the command again with --verbose.
```

Running with --verbose gives me this information:

```
   Compiling hyperfine v0.4.0
[...]

PR #32, opened 2018-02-02
Windows compatibility (WIP)

sharkdp:
Follow up for #30.

/CC @stevepentland

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,8 +13,10 @@ version = "0.4.0"
 colored = "1.6"
 indicatif = "0.9"
 statistical = "0.1"
-libc = "0.2"
 atty = "0.2.2"
+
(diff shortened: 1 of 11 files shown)


stevepentland:
I like these changes, I hadn't thought to hide the entire use inside of a function like that.

sharkdp:
Thank you for the review.

For now, this will show 0.0 as user/system time on Windows.

Maybe there is a way to measure CPU times of the child processes via [`GetProcessTimes`](https://docs.rs/winapi/*/x86_64-pc-windows-msvc/winapi/um/processthreadsapi/fn.GetProcessTimes.html) on Windows.

sharkdp:
I'm actually going to merge this, as it fixes #28. I'll open a new ticket for proper support on Windows.

stevepentland:
I’ll look into the windows performance API for a bit and see if we can fill the gap

PR #36, opened 2018-02-13
correctly estimate full execution time

### LINKED ISSUE #27: Correctly estimate full execution time if --prepare is used
The estimated execution time should include the time to run the preparation command, e.g:
```
hyperfine --prepare 'sleep 10' 'sleep 0.1'
```

rleungx:
Resolves #27.

### DIFF
--- a/src/hyperfine/benchmark.rs
+++ b/src/hyperfine/benchmark.rs
@@ -14,7 +14,7 @@ use hyperfine::timer::cputimer::CPUTimer;
 use hyperfine::shell::run_shell_command;
 
 /// Results from timing a single shell command
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Default, Copy, Clone)]
 pub struct TimingResult {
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you for working on this.

Probably I didn't explain it carefully in the attached ticket, but we definitely don't want to add the execution time of the preparation-command to the actual timing results (`times_real`, etc). The preparation command can be used to do some clean-up work and we don't want that to be included in the time measurement.

We do, however, want to measure the execution time of the preparation command once (like you did) and use it to correctly estimate `runs_in_min_time`. Otherwise, a long-running preparation command would lead to unusually long benchmark times (see the example in the attached ticket).

rleungx:
Thanks for your explaination! I misunderstood the purpose of this issue.

rleungx:
@sharkdp Please correct me if I'm still wrong.

sharkdp:
Looks great, thank you very much

PR #46, opened 2018-03-20
A few updates for the exporter code

sharkdp:
@stevepentland I would really appreciate your feedback / review.

### DIFF
--- a/src/hyperfine/export/json.rs
+++ b/src/hyperfine/export/json.rs
@@ -2,7 +2,7 @@ use super::{ExportEntry, Exporter};
 
 use std::io::{Error, ErrorKind, Result};
 
-use serde_json::to_vec_pretty;
+use serde_json::to_string_pretty;
 
 #[derive(Serialize, Debug)]
 struct HyperfineSummary<'a> {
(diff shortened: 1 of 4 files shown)


src/hyperfine/export/json.rs
```diff
@@ -13,14 +13,7 @@ pub struct JsonExporter {}
...
     fn serialize(&self, results: &Vec<ExportEntry>) -> Result<String> {
-        let serialized = to_string_pretty(&HyperfineSummary { results });
-
-        serialized.map_err(|e| {
-            Error::new(
-                ErrorKind::Other,
-                format!("Error while serializing to JSON: {:}", e),
-            )
-        })
+        to_string_pretty(&HyperfineSummary { results }).map_err(|e| Error::new(ErrorKind::Other, e))
```
stevepentland: much better, wish I had seen this in the first place

src/hyperfine/export/mod.rs
```diff
@@ -87,28 +92,29 @@ impl ExportManager {
     }
 
     /// Add an additional exporter to the ExportManager
-    pub fn add_exporter(&mut self, for_type: ExportType) {
-        self.exporters.push(for_type);
+    pub fn add_exporter(&mut self, export_type: ExportType, filename: &str) {
```
stevepentland: After seeing this, your thoughts about splitting the filename from the enum makes more sense. And this looks really good. Is there a particular benefit we get from this?

sharkdp: My thought was that we could save some duplication because we don't have to unpack the filename from each `ExportType`.

sharkdp:
Thank you very much for the review! Looking forward to the Markdown export! :smile:

PR #48, opened 2018-03-20
Export list of runtimes to JSON

sharkdp:
It's a shame that we have to introduce that `CsvEntry` struct but I didn't find a better way.

The resulting JSON file looks like this
``` json
{
  "results": [
    {
      "command": "sleep 0.2",
      "mean": 0.2043455445007143,
      "stddev": 0.0019188554787314525,
      "user": 0.0023347521428571428,
      "system": 0.002159442142857143,
      "min": 0.20007057271500004,
      "max": 0.207549326715,
      "times": [
        0.20007057271500004,
        0.20305231671500002,
        0.202839658715,
        0.205020261715,
        0.20467369571500003,
        0.20455664371500002,
        0.20604999871500002,
        0.20499965771500004,
        0.20321977971500003,
        0.20705682471500004,
        0.20308524271500004,
        0.203544831715,
        0.207549326715,
        0.20511881171500002
      ]
    },
    {
      "command": "sleep 0.3",
      "mean": 0.30501838271500006,
      "stddev": 0.001610254535237144,
      "user": 0.0036618289999999999,
      "system": 0.0015725459999999998,
      "min": 0.30210786171500006,
      "max": 0.30712019371500007,
      "times": [
        0.30562801371500006,
        0.30695781471500008,
        0.30459106071500005,
        0.30416734771500006,
        0.30712019371500007,
        0.30367873171500006,
        0.30409325871500006,
        0.305167409715,
        0.30667213471500007,
        0.30210786171500006
      ]
    }
  ]
}                                                                                                      
```

@stevepentland FYI.

### DIFF
--- a/src/hyperfine/benchmark.rs
+++ b/src/hyperfine/benchmark.rs
@@ -302,5 +302,6 @@ pub fn run_benchmark(
         system_mean,
         t_min,
         t_max,
+        times_real,
     ))
 }
--- a/src/hyperfine/export/csv.rs
+++ b/src/hyperfine/export/csv.rs
@@ -4,21 +4,6 @@ use std::io::{Error, ErrorKind, Result};
 
 use csv::WriterBuilder;
 
-use hyperfine::internal::Second;
-
-/// Set of values that will be exported.
-#[derive(Debug, Serialize)]
-pub struct CsvEntry {
-    command: String,
-    mean: Second,
-    stddev: Second,
-    user: Second,
-    system: Second,
-    min: Second,
-    max: Second,
-}
-
-
 #[derive(Default)]
 pub struct CsvExporter {}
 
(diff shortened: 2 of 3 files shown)


stevepentland:
@sharkdp is there any way we could use the skip_serializing_if that serde offers?
https://serde.rs/field-attrs.html#serdeskipserializingif--path

sharkdp:
I looked at exactly the same thing (`skip_serializing_if`), but the function that is called has the signature `fn(&T) -> bool` where `T` is the type of the field (`Vec<Second>`) - so I guess there is no way to tell if this is called from the Csv or the Json exporter (unless we would rely on a global mutable variable, which would be a hack).

stevepentland:
Yeah, that's too bad. The only other thing I could think of would be to set the field to skip if empty, and then make a copy with an empty vec. But the only benefit to that is no extra struct for CSV, all the other copying of values would still have to happen.

sharkdp:
> The only other thing I could think of would be to set the field to skip if empty, and then make a copy with an empty vec

I like that!

> all the other copying of values would still have to happen.

True, but I don't think we have to care about performance here. Hyperfine's runtime will be determined by the benchmarks themselves. Everything else is unnoticeably fast.

stevepentland:
Nice, I like the use of the Option instead, much cleaner

Issue #19, opened 2018-01-20
Parametrized benchmarks


sharkdp:
It could be interesting to run benchmarks where a (numerical) parameter is systematically changed, for example:

``` bash
> hyperfine 'make -j{}' --parameter-range 1..8
```
(modulo syntax)

sharkdp:
Another option could be to export an environment variable:
``` bash
> hyperfine --parameter num_threads 1..8 'make -j $num_threads'
```

stevepentland:
This seems like an interesting idea. I personally prefer the first style to the second, but since we'd have to parse apart the input anyway, why not just add --range-min and --range-max values to make it more explicit. For the second example, it may offer some benefit to at least offer a parameter that will set/unset environment variables during runs so that they can be used without users having to remember to manually add and remove them.

sharkdp:
> I personally prefer the first style to the second, but since we'd have to parse apart the input anyway

I think the second style would allow us to *not* parse the command. We would simply set the environment variable in the current environment and then call the command (via the shell, as usual). Since we spawn a sub-process, the environment variable would be available in the shell. This would work in a similar way to:
```
> export sleep_time=0.2
> hyperfine 'sleep $sleep_time'  # note the single quotes. $sleep_time is not expanded when we call hyperfine.
```

> why not just add --range-min and --range-max values to make it more explicit.

That sounds reasonable.

> For the second example, it may offer some benefit to at least offer a parameter that will set/unset environment variables during runs so that they can be used without users having to remember to manually add and remove them.

I'm not sure I understand.

sharkdp:
I'm currently working on this...

stevepentland:
>I'm not sure I understand.

I was just thinking that it may be beneficial to offer user's the opportunity to pass in environment variables they want to have set in the child processes we run their applications in. In many cases these could be given as part of the command, but that both makes the original command argument longer and harder for the user to quickly change between runs, and also is not as portable between platforms. On PowerShell, setting environment variables isn't quite as easy as on Linux/macOS where ```VAR=val <cmd>``` will work.

sharkdp:
> I was just thinking that it may be beneficial to offer user's the opportunity to pass in environment variables they want to have set in the child processes we run their applications in.

To be honest, I'm not sure why we should add this to hyperfine. As you said, this will work just fine when done from the shell (because the environment will be inherited by the child processes). Even in PowerShell, it would only be
``` PowerShell
$env:MyVariable = "SomeValue"
[...]

PR #51, opened 2018-03-24
Implement '--parameter-range var min max', see #19

sharkdp:
A first draft for `--parameter-range`:
```
    -r, --parameter-range <VAR> <MIN> <MAX>
            Perform benchmark runs for each value in the range MIN..MAX. Replaces the
            string '{VAR}' in each command by the current parameter value.
```

It can be used like this:

``` 
> hyperfine --parameter-range time 1 3 'sleep {time}'

Benchmark #1: sleep 1

  Time (mean ± σ):      1.002 s ±  0.001 s    [User: 0.7 ms, System: 2.0 ms]
 
  Range (min … max):    1.001 s …  1.003 s
 
Benchmark #2: sleep 2

  Time (mean ± σ):      2.003 s ±  0.002 s    [User: 1.6 ms, System: 1.8 ms]
 
  Range (min … max):    2.000 s …  2.005 s
 
Benchmark #3: sleep 3

  Time (mean ± σ):      3.002 s ±  0.002 s    [User: 1.5 ms, System: 0.9 ms]
 
  Range (min … max):    3.000 s …  3.005 s

```

@stevepentland f.y.i.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -30,6 +30,7 @@ extern crate approx;
 use std::cmp;
 use std::error::Error;
 use std::io;
+use std::ops::Range;
 
 use atty::Stream;
 use colored::*;
@@ -38,9 +39,11 @@ use clap::{App, AppSettings, Arg};
 mod hyperfine;
 
 use hyperfine::internal::write_benchmark_comparison;
-use hyperfine::types::{BenchmarkResult, CmdFailureAction, HyperfineOptions, OutputStyleOption};
+use hyperfine::types::{BenchmarkResult, CmdFailureAction, Command, HyperfineOptions,
+                       OutputStyleOption};
 use hyperfine::benchmark::{mean_shell_spawning_time, run_benchmark};
 use hyperfine::export::{ExportManager, ExportType};
+use hyperfine::error::ParameterRangeError;
 
 /// Print error message to stderr and terminate
 pub fn error(message: &str) -> ! {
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -138,6 +160,17 @@ fn main() {
                 .short("i")
                 .help("Ignore non-zero exit codes."),
         )
+        .arg(
+            Arg::with_name("parameter-range")
```
stevepentland: If we're always expecting two values we may want to defer to clap to ensure they are both there by using [```number_of_values(2)```](https://docs.rs/clap/2.31.2/clap/struct.Arg.html#method.number_of_values)

src/main.rs
```diff
@@ -212,7 +245,26 @@ fn main() {
         options.failure_action = CmdFailureAction::Ignore;
     }
 
-    let commands = matches.values_of("command").unwrap().collect();
+    let command_strings = matches.values_of("command").unwrap();
+
+    let commands = if let Some(args) = matches.values_of("parameter-range") {
+        match parse_parameter_range_args(args) {
+            Ok((param_name, param_range)) => {
```
stevepentland: Using the above may remove some need to validate the fact we were given two values

stevepentland:
It looks like the layout of the new Command struct will conflict with #53

src/main.rs
sharkdp: I added this in the beginning, but it doesn't seem to be necessary because we use `.value_names(&["VAR", "MIN", "MAX"])`, which seems to set this as well.

sharkdp:
Updated after resolving all conflicts (due to #53).

PR #62, opened 2018-05-14
[WIP] Add initial support for --shell

mqudsi:
I thought the options array would be exposed/cloned/passed-around much
more than it actually was; ended up having to pass around a shell
parameter separately to get it to where it needed to be.

What's missing is some paradigm for setting the "default" shell, since I
don't think aliasing `hyperfine` to `hyperfine -S fish` would support
`hyperfine -s sh` for a one-off non-fish benchmark, as I think clap
doesn't handle specifying a command twice to override the first
argument.

`--shell foo` or `-S foo` (`-s` is already taken)

### DIFF
--- a/src/hyperfine/app.rs
+++ b/src/hyperfine/app.rs
@@ -78,6 +78,17 @@ fn build_app() -> App<'static, 'static> {
                 .value_names(&["VAR", "MIN", "MAX"]),
         )
         .arg(
+            Arg::with_name("shell")
+                .long("shell")
+                .short("S")
+                .takes_value(true)
+                .value_name("SHELL")
+                .help(
+                    "Specifies the shell environment to set up under which the benchmarked \
+                     commands will bexe executed.",
+                )
+        )
+        .arg(
             Arg::with_name("style")
                 .long("style")
                 .short("s")
--- a/src/hyperfine/shell.rs
+++ b/src/hyperfine/shell.rs
@@ -34,10 +34,10 @@ pub fn execute_and_time(command: &str) -> io::Result<ExecuteResult> {
 
 /// Execute the given command and return a timing summary
 #[cfg(not(windows))]
-pub fn execute_and_time(command: &str) -> io::Result<ExecuteResult> {
+pub fn execute_and_time(shell: &str, command: &str) -> io::Result<ExecuteResult> {
     let cpu_timer = get_cpu_timer();
 
-    let status = run_shell_command(command)?;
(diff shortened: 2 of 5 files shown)


src/hyperfine/app.rs
```diff
@@ -77,6 +77,17 @@ fn build_app() -> App<'static, 'static> {
...
         )
+        .arg(
+            Arg::with_name("shell")
+                .long("shell")
+                .short("S")
+                .takes_value(true)
+                .value_name("SHELL")
+                .help(
+                    "Specifies the shell environment to set up under which the benchmarked \
+                     commands will bexe executed.",
```
sharkdp: Small typo.

src/hyperfine/shell.rs
```diff
@@ -62,7 +62,7 @@ fn run_shell_command(command: &str) -> io::Result<std::process::ExitStatus> {
 
 /// Run a Windows shell command using cmd.exe
 #[cfg(windows)]
-fn run_shell_command(command: &str) -> io::Result<std::process::Child> {
+fn run_shell_command(_: &str, command: &str) -> io::Result<std::process::Child> {
     Command::new("cmd")
```
sharkdp: Why should this be disabled for Windows? Granted, it would only work if the shell accepts `/C`.

src/main.rs
```diff
@@ -171,6 +172,12 @@ fn build_export_manager(matches: &ArgMatches) -> ExportManager {
     export_manager
 }
 
+/// Retrieve shell parameter or set default
+fn get_shell<'a>(matches: &'a ArgMatches) -> &'a str {
+    matches.value_of("shell")
+        .unwrap_or("sh")
```
sharkdp: This should probably be something like (?)
``` rust
let default_shell = if cfg!(unix) { "sh" } else { "cmd" };

matches.value_of("shell")
       .unwrap_or(default_shell)
```

sharkdp:
> What's missing is some paradigm for setting the "default" shell, since I
> don't think aliasing hyperfine to hyperfine -S fish would support
> hyperfine -s sh for a one-off non-fish benchmark, as I think clap
> doesn't handle specifying a command twice to override the first
> argument.

It does! We would have to configure this option to override with itself: https://docs.rs/clap/2.31.2/clap/struct.Arg.html#method.overrides_with

jasonpeacock:
I picked up this existing PR and updated it per the feedback & latest changes, it's available as #96.

sharkdp:
Closed in favor of #96.

PR #59, opened 2018-04-13
Add capture-output option

sevagh:
Not sure if this should be merged as-is. I had some errors running my command with Hyperfine (turns out it's some deep black magic errors with Numpy/Numba), but I thought it could be useful to have Hyperfine give you the stdout/stderr so you could debug a failing run.

If this is useful and/or desirable for the project, I can iterate over this PR with feedback.

PR #63, opened 2018-05-21
Allow forwarding benchmark's stdout to hyperfine

chrisduerr:
This change makes it possible to forward the stdout of the command the
user tries to benchmark to hyperfine's stdout using the `--print-stdout`
flag.

This should be irrelevant for most use-cases, however it makes it
possible to get some crude benchmarks of terminal emulators similar to
what's currently often done with `time cat file`.

I'm not quite sure if this is a desired feature for hyperfine because it's
probably not gonna be required by most users. But since I need it for my
personal usecase anyways, I thought I might as well send a PR.

### DIFF
--- a/src/hyperfine/app.rs
+++ b/src/hyperfine/app.rs
@@ -119,12 +119,13 @@ fn build_app() -> App<'static, 'static> {
                 .help("Export the timing results as a Markdown table to the given FILE."),
         )
         .arg(
-            Arg::with_name("print-stdout")
-                .long("print-stdout")
+            Arg::with_name("show-output")
+                .long("show-output")
                 .conflicts_with("style")
-                .help("Print the stdout of the benchmark instead of suppressing it. \
+                .help("Print the stdout and stderr of the benchmark instead of suppressing it. \
                        This will increase the time it takes for benchmarks to run, \
-                       so it should only be used when trying to benchmark output speed."),
+                       so it should only be used for debugging purposes or \
+                       when trying to benchmark output speed."),
         )
         .help_message("Print this help message.")
         .version_message("Show version information.")
--- a/src/hyperfine/app.rs
+++ b/src/hyperfine/app.rs
@@ -118,6 +118,14 @@ fn build_app() -> App<'static, 'static> {
                 .value_name("FILE")
                 .help("Export the timing results as a Markdown table to the given FILE."),
         )
+        .arg(
+            Arg::with_name("print-stdout")
+                .long("print-stdout")
+                .conflicts_with("style")
+                .help("Print the stdout of the benchmark instead of suppressing it. \
+                       This will increase the time it takes for benchmarks to run, \
+                       so it should only be used when trying to benchmark output speed."),
+        )
         .help_message("Print this help message.")
         .version_message("Show version information.")
 }
--- a/src/hyperfine/benchmark.rs
+++ b/src/hyperfine/benchmark.rs
@@ -39,18 +39,18 @@ fn subtract_shell_spawning_time(time: Second, shell_spawning_time: Second) -> Se
 /// Run the given shell command and measure the execution time
 pub fn time_shell_command(
     command: &Command,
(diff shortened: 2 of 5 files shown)


sharkdp:

Have you seen #59? How does this compare to that PR?

(tests currently fail on Windows)

chrisduerr:
Oh I haven't looked at other PRs before implementing this, #59 looks good and it's probably fine to just close this one.

One thing I can already see in #59 though is that instead of using `Stdio::inherit()`, it uses `Stdio::pipe()` which seems to lead to some unnecessary complexity. Using `Stdio::inherit()` would probably also make it simple enough to implement the windows part without having to test anything (there shouldn't be anything that could go wrong).

My PR just includes stdout, it seems like #59 also includes stderr. I'll close this PR and add some comments to #59.

sharkdp:
> My PR just includes stdout, it seems like #59 also includes stderr. I'll close this PR and add some comments to #59.

Fantastic, thank you very much!

If you like (and only if you find the time), you could also continue working on that other branch (I guess I would merge it to a branch and you could send a PR against that branch).

chrisduerr:
Oh I've actually looked at the other PR again and it seems to do something different. It writes the stdout after the benchmark is finished instead of piping the stdout of the application directly to the stdout of hyperfine while the benchmark is running.

I can re-open my PR and fix it up on windows if something like this would be desired, but there seems to be some overlap with #59 so it might be a better option to do implement both in a single PR.

If someone else wants to get  his hands dirty, I'm fine with people re-using parts of my code too.

sharkdp:
I think I would prefer this version over #59 if we could also include `stderr` here. Maybe we can rename the command-line option to `--show-output` or something similar.

src/hyperfine/app.rs
```diff
@@ -118,6 +118,14 @@ fn build_app() -> App<'static, 'static> {
                 .value_name("FILE")
                 .help("Export the timing results as a Markdown table to the given FILE."),
         )
+        .arg(
+            Arg::with_name("print-stdout")
+                .long("print-stdout")
+                .conflicts_with("style")
```
sharkdp: :+1:

src/hyperfine/app.rs
```diff
@@ -118,6 +118,14 @@ fn build_app() -> App<'static, 'static> {
...
+                .help("Print the stdout of the benchmark instead of suppressing it. \
+                       This will increase the time it takes for benchmarks to run, \
+                       so it should only be used when trying to benchmark output speed."),
```
sharkdp: maybe add ".. only be used for debug purposes or when trying to ..."?

src/hyperfine/benchmark.rs
```diff
@@ -86,7 +96,12 @@ pub fn mean_shell_spawning_time(style: &OutputStyleOption) -> io::Result<TimingR
 
     for _ in 0..COUNT {
         // Just run the shell without any command
-        let res = time_shell_command(&Command::new(""), CmdFailureAction::RaiseError, None);
+        let res = time_shell_command(
+            &Command::new(""),
+            print_stdout,
```
sharkdp: Do we really want to show the output of the empty shell command or just pass `false` here?

chrisduerr: I wasn't too sure about that. From the name alone I thought that `time_shell_command` would be used to deduct the shell startup time from the actual application runtime, so I wanted to keep it as similar to the actual benchmark as possible.

If that's not the case or the time difference is irrelevant (it should be pretty much 0), then I can just remove it.

sharkdp: Yeah okay, I guess that makes sense. Let's leave it in there.

src/hyperfine/app.rs
sharkdp: *debugging purposes

chrisduerr:
I think I've addressed everything you've mentioned. If you find anything else, let me know!

PR #68, opened 2018-07-04
Support parameter in preparation command

### LINKED ISSUE #67: Parameters don't get expanded in preparation commands
It seems like parameters are only substituted for the actual command, not the preparation command:

    $ hyperfine -P X 1 10 -p 'python -c "print 2**{X}"' --show-output -- ls
    Benchmark #1: ls
    
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
    NameError: name 'X' is not defined
    Error: The preparation command terminated with a non-zero exit code. Append ' || true' to the command     if you are sure that this can be ignored.

siiptuo:
Fixed #67


sharkdp:
Thank you very much!!

PR #73, opened 2018-08-25
reduce overhead of println

tathanhdinh:
This PR tries to reduce the (tiny) overhead that `println!` may make (because of [locks](https://llogiq.github.io/2017/06/01/perf-pitfalls.html) and since `stdout` is [line-buffered](https://github.com/rust-lang/rust/issues/10279)). I don't think that using a buffered writer over `stdout` will make a noticeable performance gain in this case, so feel free to reject it if you don't like.

### DIFF
--- a/src/hyperfine/benchmark.rs
+++ b/src/hyperfine/benchmark.rs
@@ -1,4 +1,4 @@
-use std::io;
+use std::io::{self, BufWriter, Write};
 use std::process::Stdio;
 
 use colored::*;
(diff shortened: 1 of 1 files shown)


sharkdp:
> I don't think that using a buffered writer over stdout will make a noticeable performance gain in this case

I also don't think so. Within `hyperfine` itself, I don't think we have to care about performance too much because a typical run of `hyperfine` will be completely dominated by the actual benchmarking time (which is on the order of seconds) whereas everything else (calculations, output) will run in a few milliseconds.

The `println!`-based code is a bit easier and clearer, so I'd like to keep it.

Anyway, thank you very much for your contribution :+1:

PR #74, opened 2018-08-25
change windows cmd interpreter

tathanhdinh:
The default command line interpreter of Windows is `cmd.exe`, moreover the order in which a file is selected to run [depends](https://stackoverflow.com/questions/605101/order-in-which-command-prompt-executes-files-with-the-same-name-a-bat-vs-a-cmd) also on its extension, for example in my machine:

```
$ echo %PATHEXT%
.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC
```
so if there is a file named `cmd.com` in %PATH% then it will be selected to run instead of `cmd.exe`.

### DIFF
--- a/src/hyperfine/shell.rs
+++ b/src/hyperfine/shell.rs
@@ -71,7 +71,7 @@ fn run_shell_command(
     stderr: Stdio,
     command: &str,
 ) -> io::Result<std::process::Child> {
-    Command::new("cmd")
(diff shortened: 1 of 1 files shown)


sharkdp:
Fixed in [v1.3.0](https://github.com/sharkdp/hyperfine/releases/tag/v1.3.0).

Issue #80, opened 2018-09-09
Add option to choose the units used for CLI & results output


jasonpeacock:
(a follow up to #71)

Currently the CLI automatically selects the units (seconds, milliseconds) based on the size of the `mean` value, but the results export is always in seconds.

Choosing the units will force for the CLI and results export units to always match, and allow users to specify the units they are most familiar with/integrate with their reporting systems.

The option could be `-u --units <Seconds|Milliseconds>`, and maybe extended to also include minutes (and hours)?

---

The units option value would be passed through to both `format::format_duration_units` for the CLI and `ExportManager::write_results` for the results export.

sharkdp:
Thank you for the feedback!

If this only affects the CLI and the Markdown export, this sounds good to me.

I'd suggest `-u --time-unit <millisecond|second|...>`.

Kaos-Industries:
I would also look forward to this, but instead of --time-unit can shorter options be preferred like `-t`? Long options are a hell of a lot of work, and I don't understand why some CLI programs insist on preferring long options before all of the short alternatives have been exhausted.

sharkdp:
I'm okay with a short option. I just suggested `-u` (for "unit", there are no other units in hyperfine) instead of `-t`.

> Long options are a hell of a lot of work, and I don't understand why some CLI programs insist on preferring long options before all of the short alternatives have been exhausted.

I typically add long command-line options to my programs because they have a self-documenting property which is great for usage in scripts.

Short options are great for interactive use of often-used features, but I tend to also be hesitant when adding new short flags because the namespace is really small and there might be better uses for a flag in the future. For rarely-used features (like hyperfines `--export-* <FILE>` flags), I think its perfectly fine to not have a short option because nobody will remember them anyway.

jasonpeacock:
working on this now, will use the suggested `-u --time-unit <millisecond|second|...>` options

sharkdp:
Released in [v1.4.0](https://github.com/sharkdp/hyperfine/releases/tag/v1.4.0).

PR #81, opened 2018-09-09
Change markdown export to use seconds, closes #71

### LINKED ISSUE #71: Compliant time units in CLI report and results export
When passing `--export-markdown` the resulted Markdown file includes information about `Mean` and `Min..Max` measured in `ms` even when the original result report printed to stdout is presented in seconds. I think it might be better if the produced report would use the same units since it's hard to read very long `ms` numbers.

Instructions for the behavior reproduction:

```
hyperfine 'sleep 5' --export-markdown results.md
```

Command line output:

```
kbobyrev@kbobyrev ~/d/m/profile> hyperfine 'sleep 5' --export-markdown results.md
Benchmark #1: sleep 5
  Time (mean ± σ):      5.002 s ±  0.000 s    [User: 1.1 ms, System: 2.0 ms]
  Range (min … max):    5.002 s …  5.003 s
```

**results.md**:

| Command | Mean [ms] | Min…Max [ms] |
|:---|---:|---:|
| `sleep 5` | 5002.3 ± 0.2 | 5002.1…5002.7

jasonpeacock:
The markdown results exporter was using milliseconds, but the other
results exporters are using the default seconds unit.

* Change the markdown results exporter to also use the default seconds
unit of `BenchmarkResult`.

*This will change the markdown results for all users from milliseconds
to seconds, a UX breaking change.*

* Add unit tests for the markdown exporter to verify the output.

Issue #80 proposes a new option to choose the units used for both the
CLI report and results export.

### DIFF
--- a/src/hyperfine/app.rs
+++ b/src/hyperfine/app.rs
@@ -55,6 +55,22 @@ fn build_app() -> App<'static, 'static> {
                 .help("Perform at least NUM runs for each command (default: 10)."),
         )
         .arg(
+            Arg::with_name("max-runs")
+                .long("max-runs")
+                .short("M")
+                .takes_value(true)
+                .value_name("NUM")
+                .help("Perform at most NUM runs for each command."),
+        )
+        .arg(
+            Arg::with_name("runs")
+                .long("runs")
+                .short("r")
+                .takes_value(true)
+                .value_name("NUM")
+                .help("Perform exactly NUM runs for each command."),
+        )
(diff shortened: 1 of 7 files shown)


sharkdp:
I think something went wrong while squashing the commits?

jasonpeacock:
hmm, yeah looks like the PR got confused when I rebased this morning against master. Will open a new PR.

PR #83, opened 2018-09-14
Report standard deviation of the speedup ratio

Shnatsel:
Report standard deviation of the speedup ratio.

Here are results for running the same Rust binary that exits immediately twice:

Before:
```
  'target/release/examples/decode' ran
    1.03x faster than 'target/release/examples/decode'
```

After:
```
  'target/release/examples/decode' ran
    1.03x ± 0.65 faster than 'target/release/examples/decode'
```

Results from a real-world benchmarking run:

```
  'flac -fd testsamples/extra/*.flac' ran
    1.56x ± 0.08 faster than 'target/release/examples/decode testsamples/extra/*.flac'
```

I'd also really like to see statistical significance estimation and getting rid of the word "faster" and all that green in case it's not, but I don't have the time to work on it right now.


Shnatsel:
I've updated the code with the message you suggested.

sharkdp:
Thank you!

PR #84, opened 2018-09-15
Change markdown export to use seconds, closes #71

### LINKED ISSUE #71: Compliant time units in CLI report and results export

jasonpeacock:
The markdown results exporter was using milliseconds, but the other
results exporters are using the default seconds unit.

Apply some refactoring so the same logic used to select units for the
CLI output can be used to select the units for the markdown exporter.

Now the units of the markdown exporter match that of the CLI.

When there are multiple benchmarks, the CLI will choose the
appropriate unit for each benchmark separately. In this case, the
markdown exporter will use the first benchmark to select the units for
all the results.

*This will change the markdown results for all users from milliseconds
to seconds, a UX breaking change.*

Add unit tests for the markdown exporter to verify the output.

Issue #80 proposes a new option to choose the units used for both the
CLI report and results export.

### DIFF
--- a/src/hyperfine/export/markdown.rs
+++ b/src/hyperfine/export/markdown.rs
@@ -1,40 +1,157 @@
 use super::Exporter;
+
+use hyperfine::format::{Unit, format_duration_value};
 use hyperfine::types::BenchmarkResult;
 
 use std::io::Result;
 
-const MULTIPLIER: f64 = 1e3;
+macro_rules! TABLE_HEADER {
+    ($unit:expr) => {
+        format!("| Command | Mean [{unit}] | Min…Max [{unit}] |\n|:---|---:|---:|\n", unit=$unit) };
+}
 
 #[derive(Default)]
 pub struct MarkdownExporter {}
 
 impl Exporter for MarkdownExporter {
     fn serialize(&self, results: &Vec<BenchmarkResult>) -> Result<Vec<u8>> {
-        let mut destination = start_table();
+        // Default to `Second`.
+        let mut unit = Unit::Second;
+
+        if !results.is_empty() {
+            // Use the first BenchmarkResult entry to determine the unit for all entries.
+            unit = format_duration_value(results[0].mean, None).1;
+        }
+
+        let mut destination = start_table(unit);
 
         for result in results {
-            add_table_row(&mut destination, result);
+            add_table_row(&mut destination, result, unit);
         }
 
         Ok(destination)
     }
 }
 
-fn start_table() -> Vec<u8> {
-    "| Command | Mean [ms] | Min…Max [ms] |\n|:---|---:|---:|\n"
-        .bytes()
-        .collect()
+fn start_table(unit: Unit) -> Vec<u8> {
+    TABLE_HEADER!(unit.short_name()).bytes().collect()
 }
-fn add_table_row(dest: &mut Vec<u8>, entry: &BenchmarkResult) {
+
+fn add_table_row(dest: &mut Vec<u8>, entry: &BenchmarkResult, unit: Unit) {
+    let mean_str = format_duration_value(entry.mean, Some(unit)).0;
+    let stddev_str = format_duration_value(entry.stddev, Some(unit)).0;
+    let min_str = format_duration_value(entry.min, Some(unit)).0;
+    let max_str = format_duration_value(entry.max, Some(unit)).0;
+
     dest.extend(
         format!(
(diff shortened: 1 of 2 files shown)


src/hyperfine/format.rs
```diff
@@ -16,11 +44,54 @@ pub fn format_duration(duration: Second, unit: Option<Unit>) -> String {
...
-        (format!("{:.3} s", duration), Unit::Second)
+    let (out_str, out_unit) = format_duration_value(duration, unit);
+
+    (format!("{} {}", out_str, out_unit.short_name()), out_unit)
+}
+
+/// Like `format_duration`, but returns the target unit as well.
+pub fn format_duration_value(duration: Second, unit: Option<Unit>) -> (String, Unit) {
+
+    // Default to `Second` until proven otherwise.
```
sharkdp: Why did you rewrite the logic of choosing the unit? It was much more compact before (and not using mutable variables).

src/hyperfine/format.rs
```diff
@@ -6,6 +6,34 @@ pub enum Unit {
...
+    /// The abbreviation of the Unit.
+    pub fn short_name(&self) -> String {
+        match *self {
+            Unit::Second => String::from("s"),
+            Unit::MilliSecond => String::from("ms"),
+        }
+    }
+
+    /// The multiplier value to convert from `from_unit` to the Unit.
+    pub fn multiplier_from(&self, from_unit: Unit) -> f64 {
```
sharkdp: I don't think that this method makes things more clear. If we want to keep something like this, I would rather use something like `Unit::conversion_factor_seconds()` (without any arguments), otherwise the number of cases grows quadratically with the number of units. We only use `multiplier_from(Second)` anyway.

sharkdp:
@jasonpeacock Ping on this

jasonpeacock:
sorry for the delay, responding now...

src/hyperfine/export/markdown.rs
```diff
@@ -1,39 +1,157 @@
...
+| `sleep 0.1` | 0.106 ± 0.002 | 0.102…0.108 |
+", TABLE_HEADER!("s"));
+
+    assert_eq!(formatted_expected, formatted);
+}
+
+/// An empty list of benchmark results will only include the table header
+/// in the markdown output, using the default `Seconds` unit.
+#[test]
+fn test_markdown_format_empty_results() {
```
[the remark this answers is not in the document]
jasonpeacock: removed.

src/hyperfine/export/markdown.rs
```diff
@@ -1,39 +1,157 @@
 use super::Exporter;
+
+use hyperfine::format::{Unit, format_duration_value};
 use hyperfine::types::BenchmarkResult;
 
 use std::io::Result;
 
-const MULTIPLIER: f64 = 1e3;
+macro_rules! TABLE_HEADER {
```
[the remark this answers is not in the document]
jasonpeacock: I had tried and that ran into some issues, I think because I was trying to use `str`, but using `String` instead it's happy.

Converted to function.

src/hyperfine/format.rs
jasonpeacock: It felt weird to hard-code the `Unit::` so much :) I tried to make it more dynamic and extensible, but I you're right that we're probably concerned about trying to support a bunch more unit types.

I've switched back to the original `if/else` logic.

src/hyperfine/format.rs
jasonpeacock: Good point, I think got here after some other changes and it made sense at the time, but looking at it now it's overkill :)

Removed, and just converting the values directly in `Unit.format`.

PR #85, opened 2018-09-15
Compact output, add `-s color` option, closes #70

### LINKED ISSUE #70: Compact output format, finer control over the output
I am running hyperfine with many different commands at once. In order to not only see the factor between the different commands at the end but also (some) of the means I suggest one, some or all of these changes to the output (switchable by a command line option):
* Do not print empty lines between mean and min-max and the headlines and so on. I can currently get quite close to what I want when I run `hyperfine ... | sed '/^ *$/d'`.
* The above mentioned  `hyperfine ... | sed '/^ *$/d'` is without color and when I run  `hyperfine -s full ... | sed '/^ *$/d'` instead I also get the interactive graph which doesn't play nicely with the pipe and sometimes messes up a line. So why not add a `color` style to `-s` which has color but is not interactive?
* All of these could be an additional value for `-s` and `-s` could be a comma separated list so that things like this would make sense: `hyperfine -s compact,color`, `hyperfine -s interactive,compact,nocolor`, `hyperfine -s full,compact`, `hyperfne -s basic,compact`...

jasonpeacock:
Strip the empty lines within the results for each benchmark, but keep
the empty line between each benchmark, and before the summary.

Add `color` style option to show color but not be interactive (no
progress bar).

### DIFF
--- a/src/hyperfine/app.rs
+++ b/src/hyperfine/app.rs
@@ -100,12 +100,13 @@ fn build_app() -> App<'static, 'static> {
                 .short("s")
                 .takes_value(true)
                 .value_name("TYPE")
-                .possible_values(&["auto", "basic", "full", "nocolor"])
+                .possible_values(&["auto", "basic", "full", "nocolor", "color"])
                 .help(
                     "Set output style type (default: auto). Set this to 'basic' to disable output \
                      coloring and interactive elements. Set it to 'full' to enable all effects \
                      even if no interactive terminal was detected. Set this to 'nocolor' to \
-                     keep the interactive output without any colors.",
+                     keep the interactive output without any colors. Set this to 'color to' keep \
+                     the colors without any interactive output.",
                 ),
         )
         .arg(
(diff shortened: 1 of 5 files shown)


jasonpeacock:
Example output (pretend it has colors):

```
Benchmark #1: sleep 0.2
  Time (mean ± σ):     205.9 ms ±   1.8 ms    [User: 1.1 ms, System: 1.2 ms]
  Range (min … max):   203.5 ms … 208.6 ms

Benchmark #2: sleep 0.1
  Time (mean ± σ):     105.9 ms ±   1.3 ms    [User: 1.1 ms, System: 1.2 ms]
  Range (min … max):   102.5 ms … 108.7 ms

Benchmark #3: sleep 0.3
  Time (mean ± σ):     306.5 ms ±   1.8 ms    [User: 1.1 ms, System: 1.3 ms]
  Range (min … max):   303.7 ms … 309.3 ms

Summary
  'sleep 0.1' ran
    1.94x faster than 'sleep 0.2'
    2.89x faster than 'sleep 0.3'
```

PR #95, opened 2018-10-30
Add option to set time units for CLI & Markdown, closes #80

### LINKED ISSUE #80: Add option to choose the units used for CLI & results output

jasonpeacock:
Add option to set the time units to either `MilliSecond` or `Second`
for both CLI output and Markdown export, overriding the default logic.

### DIFF
--- a/src/hyperfine/app.rs
+++ b/src/hyperfine/app.rs
@@ -86,13 +86,13 @@ fn build_app() -> App<'static, 'static> {
             Arg::with_name("parameter-scan")
                 .long("parameter-scan")
                 .short("P")
+                .takes_value(true)
+                .allow_hyphen_values(true)
+                .value_names(&["VAR", "MIN", "MAX"])
                 .help(
                     "Perform benchmark runs for each value in the range MIN..MAX. Replaces the \
                      string '{VAR}' in each command by the current parameter value.",
-                )
-                .takes_value(true)
-                .allow_hyphen_values(true)
-                .value_names(&["VAR", "MIN", "MAX"]),
+                ),
         )
         .arg(
             Arg::with_name("style")
@@ -123,6 +123,16 @@ fn build_app() -> App<'static, 'static> {
                 .long("ignore-failure")
                 .short("i")
                 .help("Ignore non-zero exit codes."),
+        )
+        .arg(
+            Arg::with_name("time-unit")
+                .long("time-unit")
+                .short("u")
+                .takes_value(true)
+                .value_name("millisecond|second")
+                .possible_values(&["millisecond", "second"])
+                .help( "Set the time unit used for CLI output and Markdown export.",
+                ),
         )
         .arg(
             Arg::with_name("export-csv")
--- a/src/hyperfine/app.rs
+++ b/src/hyperfine/app.rs
@@ -129,10 +129,9 @@ fn build_app() -> App<'static, 'static> {
                 .long("time-unit")
                 .short("u")
                 .takes_value(true)
-                .value_name("millisecond|second")
+                .value_name("UNIT")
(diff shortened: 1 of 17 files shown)


src/hyperfine/export/markdown.rs
```diff
@@ -1,16 +1,19 @@
...
 use std::io::Result;
 
 #[derive(Default)]
 pub struct MarkdownExporter {}
 
 impl Exporter for MarkdownExporter {
-    fn serialize(&self, results: &Vec<BenchmarkResult>) -> Result<Vec<u8>> {
-        let unit = if let Some(first_result) = results.first() {
+    fn serialize(&self, results: &Vec<BenchmarkResult>, unit: Option<Unit>) -> Result<Vec<u8>> {
+        let unit = if unit.is_some() {
```
sharkdp: Can we use `if let Some(..) = ` here to avoid the `.unwrap()`?

src/hyperfine/app.rs
```diff
@@ -115,6 +115,16 @@ fn build_app() -> App<'static, 'static> {
                 .short("i")
                 .help("Ignore non-zero exit codes."),
         )
+        .arg(
+            Arg::with_name("time-unit")
+                .long("time-unit")
+                .short("u")
+                .takes_value(true)
+                .value_name("millisecond|second")
```
sharkdp: I would suggest to just use a value name of `UNIT` here in analogy to the other options. I would list the possible values in the `.help(..)` text: "Possible values: millisecond, second.".

src/hyperfine/app.rs
```diff
@@ -115,6 +115,16 @@ fn build_app() -> App<'static, 'static> {
...
+                .possible_values(&["millisecond", "second"])
+                .help( "Set the time unit used for CLI output and Markdown export.",
```
sharkdp: I'd remove the part about "CLI output" and "Markdown output". Maybe just: "Explicitly set the unit of time. Possible values: .."?

src/main.rs
```diff
@@ -95,15 +95,15 @@ fn main() {
     let export_manager = build_export_manager(&matches);
     let commands = build_commands(&matches);
 
-    let res = match options {
-        Ok(opts) => run(&commands, &opts),
-        Err(e) => error(e.description()),
+    let res = match &options {
+        &Ok(ref opts) => run(&commands, &opts),
```
sharkdp: `ref` is ok here, but why match on a borrowed `&options` instead of just `options`?

jasonpeacock: huh...I got here via a circuitous route and looks like `ref` fixed the issue that I first tried solving with borrowing.

Fixed :)

src/hyperfine/types.rs
```diff
@@ -4,6 +4,31 @@ use std::fmt;
 /// Type alias for unit of time
 pub type Second = f64;
 
+/// Supported time units
```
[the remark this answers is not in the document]
jasonpeacock: I gave it a try, let me know what you think.

I created a `units` module, and moved both `Units` and `Second` into it, as otherwise there was a circular dependency between `types::HyperfineOptions` which needs `units::Units`, but then `units::Units` needs `types::Second`.

jasonpeacock:
Updated!

Issue #110, opened 2018-12-03
align `--output-json` with `--output-csv`


psteinb:
in version `1.4.0` of hyperfine `--output-json test.json` yields the individual timings, whereas `--output-csv test.csv` does not?

```
$ hyperfine --export-csv test.csv 'sleep 0.2'
Benchmark #1: sleep 0.2
  Time (mean ± σ):     204.6 ms ±   0.8 ms    [User: 0.2 ms, System: 1.1 ms]
  Range (min … max):   203.4 ms … 206.3 ms
 
$ cat test.csv
command,mean,stddev,user,system,min,max
sleep 0.2,0.2046418712378572,0.0008249316314930503,0.00018673571428571428,0.0011437410714285716,0.20336149459500003,0.20626203659500003
```
and the same for json

```
$ hyperfine --export-json test.json 'sleep 0.2'
Benchmark #1: sleep 0.2
  Time (mean ± σ):     204.7 ms ±   0.9 ms    [User: 0.3 ms, System: 1.0 ms]
  Range (min … max):   203.7 ms … 206.6 ms
 
$ cat test.json
{
  "results": [
    {
      "command": "sleep 0.2",
      "mean": 0.20466111425500003,
      "stddev": 0.0008757442708665281,
      "user": 0.00031423571428571424,
      "system": 0.0010083085714285716,
      "min": 0.203659674255,
      "max": 0.20661967825500002,
      "times": [
        0.204063288255,
        0.20453320125500002,
        0.204890268255,
        0.204969739255,
        0.203828842255,
        0.20639345525500002,
        0.20492674625500001,
        0.204334207255,
        0.203659674255,
        0.20429706725500002,
        0.20661967825500002,
        0.204421583255,
        0.20430772825500002,
        0.204010120255
      ]
    }
  ]
}
```

it would be cool, if the 2 would be aligned.

sharkdp:
How would you integrate the invidual timings in the CSV format? Multiple lines are for multiple benchmarks. The columns are for mean, stddev, etc. There is no third axis. We would have to include the individiual timings in a single cell, separated by a different character. I don't think that's really how CSV is meant to be used?

sharkdp:
In my view, the current CSV export is for people who want to quickly create a plot for multiple benchmarks.

The JSON export is for those that want to get all of the information. It is very easy to parse from most programming languages.

[...]

Issue #22, opened 2018-01-22
Provide finer-grained statistics


jonhoo:
It'd be nice to be able to show other statistics such as the 95th percentile or median runtime. [HdrHistogram](http://hdrhistogram.org/) can record this with relatively little overhead, and there's a pretty good official Rust implementation [here](https://crates.io/crates/hdrhistogram) (I'm one of the maintainers).

josephglanville:
p95/p99/p99.9 values ofcourse would be great but an option to output the entire histogram both in machine consumable format (json/csv) and/or graphically would be amazing.

jonhoo:
@josephglanville You should be able to do all of that with a HdrHistogram. Of course, it does pay some cost of accuracy in order to remain compact, but it's unlikely that users will notice. The exact trade-offs are also easy to control.

sharkdp:
This sounds really interesting.

Is it possible to get reasonable estimates from HdrHistogram even if we only have very few samples (like 10 time measurements)?

jonhoo:
HdrHistogram doesn't actually do any statistical analysis, it just gathers up recorded values in an efficient way. Think of it as keeping a count per bin (e.g., "there have been 3 samples in the range 1-3ms, 2 in 3-5, etc."), but in a "smart" way such that you generally always have good resolution, and the histogram doesn't grow unbounded if you have large discrepancies in the values.

jonhoo:
Which then in turn can give you whatever percentile you want, though obviously not with higher fidelity than what the number of samples provides. There will be a small inaccuracy due to the binning, but it should be marginal.

XANi:
@jonhoo The thing is that (i think) in most cases CLI testing will be used for less than < 100000 iterations, so you will be saving barely any memory while introducing inaccuracies. It might make sense if that was something that activates above certain threshold, but then 1 milion floats takes 8 MB if you wanted to just use "dumb" algorithm so I doubt it is worth it.

jonhoo:
@XANi you're totally right that with few samples the advantages of using HdrHistogram aren't as compelling. That said, the inaccuracies will also likely be very small, and HdrHistogram does present a nice interface for getting percentile values. It would also be (marginally) faster than scanning all the recorded samples after the fact to compute the percentiles.

psteinb:
I personally would love, if the `--output-*` options would store the timing results of each individual run instead of the summary statistics. This way creating histograms and such can be handled downstream by other tools/languages.

[...]

PR #112, opened 2018-12-05
refined help text of export flags to separate summary results from su…

psteinb:
…mmary+raw timings
see #110 and #22


sharkdp:
Thank you very much!

PR #126, opened 2019-01-02
Fix a typo in the command-line help

Calinou:
The quote was misplaced, so I moved it to where it should be :slightly_smiling_face:


sharkdp:
Thank you very much!

PR #129, opened 2019-01-08
Add cleanup command which is executed at the end of each benchmark

### LINKED ISSUE #91: Provide preparation command to be run between benchmarking different commands
I want to benchmark two versions of `rustc` against each other, measuring the build time of a project that however has some build dependencies, downloading and building which I do not want included in the benchmark.  However, after using one version of rustc, I need to `rm -rf target` to enable the other one to build; these are both self-compiled git versions of Rust and their detection of when to recompile because the compiler changed does not work reliably.

Unfortunately, this is not possible with one hyperfine invocation currently; instead I have to resort to calling hyperfine twice and clean the build dir in between. It would be great if hyperfine had a preparation command that was run in-between different commands being benchmarked, but NOT in-between different runs of the same benchmark.

colinwahl:
Fixes #91. 

I am still a beginner at Rust, so I'm very open to feedback. Thanks!

### DIFF
--- a/src/hyperfine/app.rs
+++ b/src/hyperfine/app.rs
@@ -83,6 +83,17 @@ fn build_app() -> App<'static, 'static> {
                 ),
         )
         .arg(
+            Arg::with_name("cleanup")
+                .long("cleanup")
+                .short("c")
+                .takes_value(true)
+                .value_name("CMD")
+                .help(
+                    "Execute CMD after each benchmark run. This is useful for \
+                     cleaning up resources allocated during prepare.",
+                ),
+        )
+        .arg(
             Arg::with_name("parameter-scan")
                 .long("parameter-scan")
                 .short("P")
--- a/src/hyperfine/app.rs
+++ b/src/hyperfine/app.rs
@@ -89,8 +89,10 @@ fn build_app() -> App<'static, 'static> {
                 .takes_value(true)
                 .value_name("CMD")
                 .help(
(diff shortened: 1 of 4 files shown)


sharkdp:

I will do a full review soon, but what I've seen so far looks great. I think we should work a little bit on the `--help` text to clearly explain what the difference from `--prepare` is. The prepare-command will be executed before *each* benchmarking run (several times for every benchmarked command) while the cleanup command will only be called once after a set of benchmark runs (once for every benchmarked command).

src/hyperfine/app.rs
```diff
@@ -82,6 +82,17 @@ fn build_app() -> App<'static, 'static> {
...
                 ),
         )
+        .arg(
+            Arg::with_name("cleanup")
+                .long("cleanup")
+                .short("c")
+                .takes_value(true)
+                .value_name("CMD")
+                .help(
+                    "Execute CMD after each benchmark run. This is useful for \
```
sharkdp: As stated previously, we should clarify the help text here.

colinwahl: I agree it needs some clarification, how does this sound?

"Execute CMD after the benchmarking runs for each individual command to be benchmarked.  This is useful if the commands to be benchmarked produce artifacts that need to be cleaned up."

sharkdp: Sounds great!

colinwahl: Ok! I have updated the code.

PR #131, opened 2019-01-16
Serialize command parameters to data output files

bbannier:
This patch adds any parameter used to data output files (currently JSON
and CSV files). With that users have direct access to both input values
(i.e., benchmark parameters) and the resulting timings which should
simplify follow-up analyses.

### DIFF
--- a/src/hyperfine/benchmark.rs
+++ b/src/hyperfine/benchmark.rs
@@ -412,5 +412,6 @@ pub fn run_benchmark(
         t_min,
         t_max,
         times_real,
+        cmd.get_parameter().map(|p| p.1),
     ))
 }
--- a/src/hyperfine/types.rs
+++ b/src/hyperfine/types.rs
@@ -173,6 +173,10 @@ pub struct BenchmarkResult {
     /// All run time measurements
     #[serde(skip_serializing_if = "Option::is_none")]
     pub times: Option<Vec<Second>>,
+
+    /// Any parameter used
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub parameter: Option<i32>,
 }
 
 impl BenchmarkResult {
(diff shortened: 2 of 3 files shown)


sharkdp:
Thank you very much for your contribution.

Can you please provide a little bit of background / motivation for this change? This would make it easier to review.

bbannier:
@sharkdp, thank you for responding, and sorry for the sloppy PR.

I have updated the initial comment and the subject. Ultimately this patch is driven by me wanting to directly plot results from JSON output files for parameterized benchmarks (e.g., to inspect scaling behavior identify breakdown of naive scaling). For that I need direct access to benchmark parameter which currently requires command string parsing or adding values by hand.

I also realized that in addition to JSON, CSV output could also benefit from this improvement and updated the patches for that (which actually lead to a simplification).

src/hyperfine/benchmark.rs
```diff
@@ -412,5 +412,9 @@ pub fn run_benchmark(
         t_min,
         t_max,
         times_real,
+        match cmd.get_parameter() {
```
sharkdp: Could this `match` statement be shortened to something like
``` rust
cmd.get_parameter().map(|p| p.1)
```
?

src/hyperfine/types.rs
```diff
@@ -196,6 +201,7 @@ impl BenchmarkResult {
             min,
             max,
             times: Some(times),
+            parameter: parameter,
```
sharkdp: I think this could just be `parameter` instead of `parameter: parameter`.

sharkdp:
Thank you for the updates!

sharkdp:
Released in [v1.6.0](https://github.com/sharkdp/hyperfine/releases/tag/v1.6.0).

PR #137, opened 2019-01-29
Add Asciidoc export option

5paceToast:
This adds an option to export to an asciidoc-style table.
Functionally similar to the current markdown exporter.

### DIFF
--- a/src/hyperfine/app.rs
+++ b/src/hyperfine/app.rs
@@ -147,6 +147,13 @@ fn build_app() -> App<'static, 'static> {
                 .help("Set the time unit used. Possible values: millisecond, second."),
         )
         .arg(
+            Arg::with_name("export-asciidoc")
+                .long("export-asciidoc")
+                .takes_value(true)
+                .value_name("FILE")
+                .help("Export the timing summary statistics as an Asciidoc table to the given FILE."),
+        )
+        .arg(
             Arg::with_name("export-csv")
                 .long("export-csv")
                 .takes_value(true)
--- a/src/hyperfine/export/asciidoc.rs
+++ b/src/hyperfine/export/asciidoc.rs
@@ -1,0 +1,207 @@
+use super::Exporter;
+
+use hyperfine::format::format_duration_value;
+use hyperfine::types::BenchmarkResult;
+use hyperfine::units::Unit;
+
+use std::io::Result;
+
+#[derive(Default)]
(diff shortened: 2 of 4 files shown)


sharkdp:

Can you please show an exemplary output here in the description and give a short explanation why we would need an AsciiDoc export option in addition to the Markdown export option (I am not familar with AsciiDoc).

5paceToast:
Absolutely!

`hyperfine --export-asciidoc /dev/tty 'sleep 0.1' 'sleep 0.2'` will currently output the following:
```adoc
[cols="<,>,>"]
|===
| Command | Mean [ms] | Min…Max [ms]

| `sleep 0.1`
| 107.2 ± 2.2
| 100.4…109.2

| `sleep 0.2`
| 206.7 ± 2.6
| 200.9…209.1
|===
```
[Here](https://i.imgur.com/xaxvru5.png) is an image of what that looks like under default CSS settings, as per asciidoctorjs.

As to what asciidoc is, and why it should be used...

AsciiDoc is another markup language (similar to Markdown).
It has a *lot* more functionality behind it (it should be roughly feature-compatible with docbook - its origin).
The current canonical implementation is [AsciiDoctor](https://asciidoctor.org/), you can take a look at the plethora of features in the [user manual](https://asciidoctor.org/docs/user-manual/).
The primary issue with markdown is that it alone is generally not sufficient - and thus there are many extended variants (github markdown, gitlab markdown, commonmark, etc), whereas asciidoc is relatively complete (and currently being standardized), making it a nicer common platform, with sufficient features for most other uses (I use it for technical documentation, writing manual pages, research papers, bookwriting, and so on).
Or, shorter - Markdown is a quick but useful tool for getting *some* text into html.
Asciidoc is a comprehensive content markup, with many available targets, and useful for all sorts of use-cases.

As for the technical side, asciidoc tables look/work differently from markdown ones ([user manual section on tables](https://asciidoctor.org/docs/user-manual/#tables)).

I'm currently running some expansive benchmarks on common unix compressors, and the content will be too complex for markdown (and I tend to write in asciidoc to begin with).
While it's too late for me to restart the (very long) process now, this patch will mean I won't need to construct the tables by hand next time.

sharkdp:
Released in [v1.6.0](https://github.com/sharkdp/hyperfine/releases/tag/v1.6.0).

PR #179, opened 2019-06-19
Enable link-time optimization for release builds

Calinou:
This makes binaries slightly smaller (about 230 KB less for a 64-bit Linux binary), at the cost of longer build times.


sharkdp:
Thank you!

PR #194, opened 2019-07-17
Fix some lints

phimuemue:
I took the liberty to run `cargo clippy` on this project and address some issues.

I.e. this does/should not change the observable behaviour of the program in any way.


sharkdp:
Thank you! I wouldn't always apply clippy suggestions blindly, but all of these look really sensible.

btw: some of the code couldn't be written like this in an older version of Rust.

Issue #184, opened 2019-06-29
Add possibility to pass float values or range to --parameter-scan


guilhermeleobas:
When using hyperfine, I am missing the possibility of specifying a range or floating point values on `--parameter-scan`:

```
hyperfine --parameter-scan i 1 10 2 "sleep {i}"
```

or 

```
hyperfine --parameter-scan i 0.1 1.0 "sleep {i}"
```

sharkdp:
Thank you for the feedback. What CLI would you propose for this? We can't let `--parameter-scan` take an optional fourth argument.

guilhermeleobas:
I was thinking in use `--parameter-scan` with an optional fourth argument but since that is not possible, maybe use `--parameter-range-scan` with the same semantics of the former but with a parameter indicating the range increment.

sharkdp:
Another option would be to keep `--parameter-scan` as it is, and add a new `--parameter-step-size <step>` (or `--step-size <step>`) option.

sharkdp:
> edit2: Or instead of adding a floating point option, add a way to format the variable generate by `--parameter-scan` using the printf notation.

You can already do this - sort of.

With `bc`:
```bash
hyperfine -P var 1 5 'sleep $(echo "{var}/10" | bc -l)'
```

With `zsh`:
``` bash
hyperfine -S zsh -P var 1 5 'sleep $(({var}.0/10))'
```

piyushrungta25:
@sharkdp Opened #203 to add `--parameter-step-size` argument.

sharkdp:
@piyushrungta25 has added support for `--parameter-step-size` in #203. I think it would be great if we could also add support for floating point values, such that we can run things like:

``` bash
hyperfine --parameter-scan var 0.0 1.6 --parameter-step-size 0.1 …
```

piyushrungta25:
> I think it would be great if we could also add support for floating point values

I see two issues with this.

1. Choosing default value of the step. Something like `0.1` as the default doesn't seem very intuitive. One thing we can do is make the step size argument required for ranges with float bounds.

2. The range iterators in std library(or even [`num`](https://docs.rs/num-iter/0.1.37/num_iter/index.html) crate) does not take float parameters. So if we are okay with rolling our own iterator, I can come up with a solution.

sharkdp:
>     1. Choosing default value of the step. Something like `0.1` as the default doesn't seem very intuitive. One thing we can do is make the step size argument required for ranges with float bounds.

[...]

PR #203, opened 2019-08-21
Add argument to take parameter range step size

piyushrungta25:
#184


sharkdp:
Thank you for your contribution, this looks great!

PR #206, opened 2019-08-31
Add support for floats in --parameter-scan

piyushrungta25:
#184

### DIFF
--- a/src/hyperfine/error.rs
+++ b/src/hyperfine/error.rs
@@ -39,7 +39,11 @@ impl Error for ParameterScanError {
             ParameterScanError::EmptyRange => "Empty parameter range",
             ParameterScanError::TooLarge => "Parameter range is too large",
             ParameterScanError::ZeroStep => "Zero is not a valid parameter step",
-            ParameterScanError::StepRequired => "Step is required when range bounds are floats",
+            ParameterScanError::StepRequired => {
+                "A step size is required when the range bounds are \
+                 floating point numbers. The step size can be specified \
+                 with the '--parameter-step-size' parameter"
+            }
         }
     }
 }
--- a/src/hyperfine/error.rs
+++ b/src/hyperfine/error.rs
@@ -1,3 +1,4 @@
+use rust_decimal::Error as DecimalError;
 use std::error::Error;
 use std::fmt;
 use std::num;
@@ -5,14 +6,22 @@ use std::num;
 #[derive(Debug)]
 pub enum ParameterScanError {
     ParseIntError(num::ParseIntError),
+    ParseDecimalError(DecimalError),
     EmptyRange,
     TooLarge,
     ZeroStep,
+    StepRequired,
 }
 
 impl From<num::ParseIntError> for ParameterScanError {
     fn from(e: num::ParseIntError) -> ParameterScanError {
         ParameterScanError::ParseIntError(e)
+    }
+}
+
+impl From<DecimalError> for ParameterScanError {
+    fn from(e: DecimalError) -> ParameterScanError {
+        ParameterScanError::ParseDecimalError(e)
     }
 }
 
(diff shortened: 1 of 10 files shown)


piyushrungta25:
Okay so this got a little more involved than I was expecting. Let me know if there is a simpler way to do this.

src/hyperfine/error.rs
```diff
@@ -26,9 +35,11 @@ impl Error for ParameterScanError {
     fn description(&self) -> &str {
         match *self {
             ParameterScanError::ParseIntError(ref e) => e.description(),
+            ParameterScanError::ParseDecimalError(ref e) => e.description(),
             ParameterScanError::EmptyRange => "Empty parameter range",
             ParameterScanError::TooLarge => "Parameter range is too large",
             ParameterScanError::ZeroStep => "Zero is not a valid parameter step",
+            ParameterScanError::StepRequired => "Step is required when range bounds are floats",
```
sharkdp: Maybe: *"A step size is required when the range bounds are floating point numbers. The step size can be specified with the '--parameter-step-size' parameter"*

src/hyperfine/parameter_range.rs
```diff
@@ -0,0 +1,176 @@
...
+    step: Option<&str>,
+) -> Result<Vec<Command<'a>>, ParameterScanError> {
+    let command_strings = command_strings.collect::<Vec<&str>>();
+    let param_name = vals.next().unwrap();
+    let param_min = vals.next().unwrap();
+    let param_max = vals.next().unwrap();
+
+    // attempt to parse as integers
+    if let (Ok(param_min), Ok(param_max)) = (param_min.parse::<i32>(), param_max.parse::<i32>()) {
+        let step = step.unwrap_or("1").parse::<i32>()?;
```
sharkdp: This will result in an error if we run
```
--parameter-scan t 0 1 --parameter-step-size 0.1
```

I think we could potentially support this if we add the step-parsing in the `if let` above. This way, all of the parameters would be parsed as decimals below. What do you think?

sharkdp:
The CI tests for Rust 1.31 failed because you probably used some language constructs which are not supported in 1.31. That's completely fine. We just need to bump the minimum-required Rust version in `.travis.yml` (and in the README, where I forgot to update it last time)

src/hyperfine/error.rs
piyushrungta25: This message is much better than the last. I have updated it.

src/hyperfine/parameter_range.rs
piyushrungta25: This is a nice catch. I took your suggestion and moved the step parsing line in the `if let` statement.

piyushrungta25:
> Maybe we could rename the value from STEP to DELTA 

DELTA sounds great. I have added the `-D` option.

piyushrungta25:
I should also note a couple of behavior after these changes which are not immediately apparent.

- We will never see any int parsing errors. So, while previously invocation of `--parameter-scan var 1 10000000000` will show error of `number too large to fit in target type`, now it will ask for step since 10000000000 is a valid float.

- JSON report will serialize decimals as `"0.1"` instead of `0.1`. Since `rust_decimal` [explicitly serializes it as string](https://github.com/paupino/rust-decimal/blob/master/src/serde_types.rs#L69), I didn't override the behavior.

sharkdp:
Thank you for pointing this out!

>     * We will never see any int parsing errors. So, while previously invocation of `--parameter-scan var 1 10000000000` will show error of `number too large to fit in target type`, now it will ask for step since 10000000000 is a valid float.

I guess that's a good thing? :smile: 

>     * JSON report will serialize decimals as `"0.1"` instead of `0.1`. Since `rust_decimal` [explicitly serializes it as string](https://github.com/paupino/rust-decimal/blob/master/src/serde_types.rs#L69), I didn't override the behavior.

Okay. Makes sense, given that it's not actually a floating point type. Sounds acceptable to me.

piyushrungta25:
> I guess that's a good thing?

I guess. To explain my concern a little better, when you are used to invocations like `hyperfine -P var 0 10 'sleep {var}'`, it's not immediately obvious why `hyperfine -P var 0 10000000000 'sleep {var}'` is suddenly asking for a step size.
But it shouldn't be too much of a concern. I just though I should point it out.
Also thanks for the review. I love this and the other tools you wrote, especially bat and hexyl.

sharkdp:
> I guess. To explain my concern a little better, when you are used to invocations like `hyperfine -P var 0 10 'sleep {var}'`, it's not immediately obvious why `hyperfine -P var 0 10000000000 'sleep {var}'` is suddenly asking for a step size.

Right, agreed. On the other hand, 10000000000 is probably not going to appear normally.

>  Also thanks for the review. I love this and the other tools you wrote, especially bat and hexyl.

Thank you for the feedback!

Issue #216, opened 2019-09-26
Prepare commands that differ, and only run once, depending on the command being timed?


klundberg:
I have a specific request: We're using hyperfine to measure the difference in compile times between two branches of our codebase. Right now I have two separate hyperfine invocations in a script, and before each of them I check out the specific branch in git that I want to measure compile times for.

I noticed today that hyperfine supports timing and comparing multiple commands in a single invocation and that it provides stats comparing them which is great! But in order for us to use that, we'd need to do a git checkout before each run of the build command, (something like `hyperfine "git checkout branch1; build.sh" "git checkout branch2; build.sh") and I don't want that checkout time to be included in our metrics. Would it be possible to have some sort of global prepare statement per-command that's being run that runs only once, and is different for each command being tested? I imagine that would be a new feature and a nontrivial one at that.

Or would you recommend running a few warmup runs with the checkout in the command to benchmark, or something like that?

Thanks!

sharkdp:
Thank you for the feedback. That sounds like a useful feature to have.

Did you see the `--prepare <cmd>` option? This is actually pretty close to what you want, except you can only specify a single command for all benchmark runs:
```
hyperfine --prepare "git checkout ???" …
```

We could change that and allow `--prepare` to be specified either once or exactly N times (where N is the number of benchmarked commands). In the latter case, we would run command-specific cleanup commands. This way, you could use:
``` bash
hyperfine --prepare "git checkout branch1" "./build.sh" --prepare "git checkout branch2" "./build.sh"
```
What do you think?

> Or would you recommend running a few warmup runs with the checkout in the command to benchmark, or something like that?

If the `git checkout branchX` time is negligible compared to the `build.sh` time (in cases where `branchX` is already checked out!), you can definitely use `--warmup`:
``` bash
hyperfine --warmup 1 "git checkout branch1; build.sh" "git checkout branch2; build.sh"
```
The only problem is that it will perform one unnecessary build for each branch.

sharkdp:
Another thing that comes really close is the `--parameter-scan <VAR> <MIN> <MAX>` option.

Imagine for a moment that your branches would actually be called `branch1` and `branch2`. We could then run:
```
hyperfine \
  --parameter-scan number 1 2 \
  --prepare "git checkout branch{number}" \
  "build.sh"
```

So another option would be to add a new option that would allow non-numeric parameter runs. Maybe `--parameter-list <VAR> <VALUES>` which could be used like this:
```
hyperfine \
[...]

PR #218, opened 2019-10-02
Add multiple --prepare support

iamsauravsharma:
Add support for multiple --prepare option. Corresponding --prepare option will be used for command for benchmarking according to position. If --prepare option is less than benchmark command then all other remaining benchmark command will use last prepare option ( This logic may not be best logic to implement out). If prepare option is more than command for benchmarking then extra --prepare option will not be used
Part of #216

### DIFF
--- a/src/hyperfine/app.rs
+++ b/src/hyperfine/app.rs
@@ -76,10 +76,14 @@ fn build_app() -> App<'static, 'static> {
                 .long("prepare")
                 .short("p")
                 .takes_value(true)
+                .multiple(true)
+                .number_of_values(1)
                 .value_name("CMD")
                 .help(
-                    "Execute CMD before each timing run. This is useful for \
-                     clearing disk caches, for example.",
+                    "Execute CMD before each timing run. For example, this is \
+                     useful for clearing disk caches. This can be mentioned \
+                     out multiple time for each command or only one time for \
+                     all command.",
                 ),
         )
         .arg(
--- a/src/hyperfine/app.rs
+++ b/src/hyperfine/app.rs
@@ -80,10 +80,11 @@ fn build_app() -> App<'static, 'static> {
                 .number_of_values(1)
                 .value_name("CMD")
                 .help(
-                    "Execute CMD before each timing run. For example, this is \
-                     useful for clearing disk caches. This can be mentioned \
-                     out multiple time for each command or only one time for \
-                     all command.",
+                    "Execute CMD before each timing run. This is useful for \
+                     clearing disk caches, for example. \nThe --prepare option can \
+                     be specified once for all commands or multiple times, once for \
+                     each command. In the latter case, each preparation command will \
+                     be run prior to the corresponding benchmark command.",
                 ),
         )
         .arg(
--- a/src/hyperfine/app.rs
+++ b/src/hyperfine/app.rs
@@ -81,7 +81,7 @@ fn build_app() -> App<'static, 'static> {
                 .value_name("CMD")
                 .help(
                     "Execute CMD before each timing run. This is useful for \
-                     clearing disk caches, for example. \nThe --prepare option can \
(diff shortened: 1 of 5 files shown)


sharkdp:

> If --prepare option is less than benchmark command then all other remaining benchmark command will use last prepare option ( This logic may not be best logic to implement out). If prepare option is more than command for benchmarking then extra --prepare option will not be used

I would solve both problems by issuing an error message if `--prepare` has not been specified exactly once or exactly *N* times. This way, we don't have to make these decisions. Either, we use one preparation-command for all benchmark-commands, or we have a 1:1 mapping of preparation-commands to benchmark-commands.

iamsauravsharma:
@sharkdp How would you like to raise error currently I am directly raising error by calling error function which is simple. And any suggestion over error message which was raised I think this is not best error message for this error

src/hyperfine/app.rs
```diff
@@ -76,10 +76,14 @@ fn build_app() -> App<'static, 'static> {
...
+                .multiple(true)
+                .number_of_values(1)
                 .value_name("CMD")
                 .help(
-                    "Execute CMD before each timing run. This is useful for \
-                     clearing disk caches, for example.",
+                    "Execute CMD before each timing run. For example, this is \
+                     useful for clearing disk caches. This can be mentioned \
+                     out multiple time for each command or only one time for \
+                     all command.",
```
sharkdp: Can we please rephrase this to:

Execute CMD before each timing run. This is useful for clearing disk caches, for example.
The `--prepare` option can be specified once for all commands or multiple times, once
for each command. In the latter case, each preparation command will be run prior to the
corresponding benchmark command.

src/main.rs
```diff
@@ -113,7 +123,11 @@ fn build_hyperfine_options(matches: &ArgMatches<'_>) -> Result<HyperfineOptions,
         (None, None) => {}
     };
 
-    options.preparation_command = matches.value_of("prepare").map(String::from);
+    if let Some(values) = matches.values_of("prepare") {
+        options.preparation_command = Some(values.map(String::from).collect::<Vec<String>>());
+    } else {
+        options.preparation_command = None;
+    }
```
sharkdp: I think this could be written as (untested):
``` rust
options.preparation_command = matches.values_of("prepare").map(|values| {
  values.map(String::from).collect::<Vec<_>>()
});
```

src/hyperfine/benchmark.rs
```diff
@@ -234,13 +234,14 @@ pub fn run_benchmark(
...
-            Some((param, value)) => Command::new_parametrized(preparation_command, param, value),
-            None => Command::new(preparation_command),
-        });
+    let mut prepare_cmd = None;
+    if let Some(values) = &options.preparation_command {
+        if values.len() == 1 {
+            prepare_cmd = Some(new_command_from_get_parameter(cmd, &values[0]));
+        } else {
+            prepare_cmd = Some(new_command_from_get_parameter(cmd, &values[num]));
+        }
```
sharkdp: how about first storing `&values[0]` or `&values[num]` in a local variable. Then, you only need to call `new_command_from…` once. If this works, we could remove that function again and inline it here.

src/hyperfine/benchmark.rs
```diff
@@ -234,13 +234,14 @@ pub fn run_benchmark(
...
 
     // Run init command
-    let prepare_cmd = options
-        .preparation_command
-        .as_ref()
-        .map(|preparation_command| match cmd.get_parameter() {
-            Some((param, value)) => Command::new_parametrized(preparation_command, param, value),
-            None => Command::new(preparation_command),
-        });
+    let mut prepare_cmd = None;
```
sharkdp: instead of using `mut` here, I would prefer
``` rust
let prepare_cmd = if … {
  …
} else {
  None
};
```
or (even better) a `.map` over the `options.preparation_command`, like it was before.

PR #219, opened 2019-10-02
Add support to run preparation command only once while benchmarking

iamsauravsharma:
Add support to run preparation command only once. This PR adds out new option `-o/--prepare-once` option which causes to run preparation command only one time instead of running multiple time every time before running benchmark command which is current unchangeable options.

Part of #216

### DIFF
--- a/src/hyperfine/app.rs
+++ b/src/hyperfine/app.rs
@@ -83,6 +83,16 @@ fn build_app() -> App<'static, 'static> {
                 ),
         )
         .arg(
+            Arg::with_name("prepare-once")
+                .long("prepare-once")
+                .short("o")
+                .requires("prepare")
+                .help(
+                    "Execute prepare command only once initially for benchmarking. \
+                    Required to have --prepare option specified as well"
+                ),
+        )
+        .arg(
             Arg::with_name("cleanup")
                 .long("cleanup")
                 .short("c")
--- a/src/hyperfine/benchmark.rs
+++ b/src/hyperfine/benchmark.rs
@@ -253,9 +253,16 @@ pub fn run_benchmark(
     )?;
 
     // Determine number of benchmark runs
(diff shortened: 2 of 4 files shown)


sharkdp:

I don't really think that this is a part of #216. 

I also don't think that we should add this option to `hyperfine`. If someone wants to run a preparation command exactly once, it can simply be executed before running `hyperfine`:
```
preparation-cmd; hyperfine …
```

iamsauravsharma:
Currently this is not issue for hyperfine but what about after #218 merge. I think your solution will not work if each command have different preparation command.

sharkdp:
> Currently this is not issue for hyperfine but what about after #218 merge. I think your solution will not work if each command have different preparation command.

I see your point, thank you. However, I'm not sure if this is a use-case that we really have to support. This adds another command-line option and makes it even more difficult for users to comprehend the whole benchmarking process.

I think I would like to avoid adding another command-line option and more code complexity for this niche use case.

sharkdp:
Closing this for now. We might come back to this later.

jest:
@sharkdp I'm also interested in this, as running `prepare` before each `warmup` really destroys the warmup effect for me (prepare: starting up a server with different {parameters}, warmup: well... warm up the server :) ). Also, when to run a cleanup is also debatable.

Do you mind if I open an issue for discussion of the workflow and the phases of running benchmarks? I know you want to keep it simple, but I feel that there are conflicting viewpoints and maybe describing the whole would lead to the solution satisfying all needs.

sharkdp:
> Do you mind if I open an issue for discussion of the workflow and the phases of running benchmarks? I know you want to keep it simple, but I feel that there are conflicting viewpoints and maybe describing the whole would lead to the solution satisfying all needs.

No, that sounds great. Thank you.

jest:
@sharkdp Done in #247

PR #222, opened 2019-10-07
scripts: add Welch's T test

### LINKED ISSUE #217: Calculates whether the distributions of two benchmarks are different
Suppose we have 2 benchmarks of the same functionality in different implementations and want to know whether these distributions are differed or statistically the same.

Possible solution:

### Welch's t-test

Here is the TypeScript reference implementation that consumes the JSON output of hyperfine and calculates the probability.

```js
import * as fs from 'fs';

const sum = (x: number[]) => x.reduce((a: number, b: number) => a + b);
const sqsum = (x: number[], mu: number) =>
  x.reduce((a: number, b: number) => a + (b - mu) ** 2);
const cdf = (x: number, c: number) => Math.exp(-(x ** -c));

function ttest(X: number[], Y: number[]) {
  const Xn = X.length;
  const Yn = Y.length;
  const X_mu = sum(X) / Xn;
  const Y_mu = sum(Y) / Yn;
  const X_sigma = sqsum(X, X_mu) / (Xn - 1);
  const Y_sigma = sqsum(Y, Y_mu) / (Yn - 1);
  const t = (X_mu - Y_mu) / Math.sqrt(X_sigma / Xn + Y_sigma / Yn);
  console.log(t);

  return cdf(-Math.abs(t), Xn + Yn - 2) * 2.0;
}

const result = JSON.parse(fs.readFileSync('./hyperfine.json').toString());
const X = result.results[0].times;
const Y = result.results[1].times;
const p = ttest(X, Y);
console.log(p < 0.05);
```

uetchy:
Add a Python script that consumes the JSON output of hyperfine and calculates the probability of welch's t test.

Can edit PR freely.

Fixes #217

### DIFF
--- a/scripts/welch_ttest.py
+++ b/scripts/welch_ttest.py
@@ -1,0 +1,28 @@
+#!/usr/bin/python
+
+import argparse
+import json
+from scipy import stats
+
+parser = argparse.ArgumentParser()
+parser.add_argument("file", help="JSON file with benchmark results")
+args = parser.parse_args()
+
+with open(args.file) as f:
+    results = json.load(f)["results"]
+
+a, b = [x["command"] for x in results[:2]]
+X, Y = [x["times"] for x in results[:2]]
+
+print("Command 1: {}".format(a))
+print("Command 2: {}\n".format(b))
+
+t, p = stats.ttest_ind(X, Y, equal_var=False)
+th = 0.05
+dispose = p < th
+print("t = {}, p = {}".format(t, p))
+print("p < {} = {}\n".format(th, dispose))
+if dispose:
+    print("There is a difference between the two benchmarks.")
+else:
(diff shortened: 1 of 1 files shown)


sharkdp:
Very cool, I love it!

I made a couple of small changes. The output looks like this:

```
▶ python welch_ttest.py almost-the-same.json 
Command 1: sleep 0.1000000
Command 2: sleep 0.1000001

t = 0.2, p = 0.842

The two benchmarks are almost the same (p >= 0.05).
```

```
▶ python welch_ttest.py slightly-different.json 
Command 1: sleep 0.100
Command 2: sleep 0.101

t = -4.33, p = 6.75e-05

There is a difference between the two benchmarks (p < 0.05).
```

Issue #193, opened 2019-07-17
[Feature request] Option to disable the default output entirely


gmolau:
I think it would be nice to be able to disable the standard output (with the progress bar) entirely. This would make it possible to cleanly pipe the JSON output to another process (via `--export-json /dev/stdout`). It seems like making the `--style` option accept `none` would be a good way to implement this. I'm unfortunately not familiar with Rust (otherwise I would just submit a PR), but maybe this could be considered for the future.

sharkdp:
Great idea, thank you very much.

In the meantime, you can use process substitution to pipe the output to another process. Not as beautiful as it could be, I agree:
```bash
> hyperfine 'sleep 0.1' --export-json >(jq '.results[0].mean') >/dev/null
0.10603645827285715
```

knidarkness:
Hi @sharkdp, I am currently using hyperfine and find it quite nice tool. If this issue is still relevant I could implement it. Is it?

sharkdp:
It is, yes. 

That would be great

knidarkness:
Great, could you assign it to me and I`ll check it on the weekend?

sharkdp:
closed via #224 by @knidarkness

sharkdp:
This works now:
```bash
hyperfine --style=none --export-json /dev/stdout 'sleep 0.1' 'sleep 0.2' | jq '.results[].mean'
```

PR #224, opened 2019-10-12
Option to disable the default output entirely [#193]

knidarkness:
This PR addresses the #193 Feature request. It allows to pass `none` as a value for `--style` option. 

If it is used, then no output (unless they are errors).

@sharkdp , may be there is some sense to print warnings as well. What do you think?

### DIFF
--- a/src/hyperfine/benchmark.rs
+++ b/src/hyperfine/benchmark.rs
@@ -5,6 +5,8 @@ use std::process::Stdio;
 use colored::*;
 use statistical::{mean, median, standard_deviation};
 
+use indicatif::{ProgressBar};
+
 use crate::hyperfine::format::{format_duration, format_duration_unit};
 use crate::hyperfine::internal::{get_progress_bar, max, min, MIN_EXECUTION_TIME};
 use crate::hyperfine::outlier_detection::{modified_zscores, OUTLIER_THRESHOLD};
@@ -94,7 +96,10 @@ pub fn mean_shell_spawning_time(
     show_output: bool,
 ) -> io::Result<TimingResult> {
     const COUNT: u64 = 200;
-    let progress_bar = get_progress_bar(COUNT, "Measuring shell spawning time", style);
+    let mut progress_bar: Option<ProgressBar> = None;
+    if style != OutputStyleOption::None { 
+        progress_bar = Some(get_progress_bar(COUNT, "Measuring shell spawning time", style)); 
+    }
 
     let mut times_real: Vec<Second> = vec![];
     let mut times_user: Vec<Second> = vec![];
@@ -132,10 +137,10 @@ pub fn mean_shell_spawning_time(
                 times_user.push(r.time_user);
                 times_system.push(r.time_system);
             }
-        }
-        progress_bar.inc(1);
+        }  
+        // if let Some(bar) = progress_bar { bar.inc(1); }
     }
-    progress_bar.finish_and_clear();
+    if let Some(bar) = progress_bar { bar.finish_and_clear(); }
 
     Ok(TimingResult {
         time_real: mean(&times_real),
@@ -193,12 +198,14 @@ pub fn run_benchmark(
     shell_spawning_time: TimingResult,
     options: &HyperfineOptions,
 ) -> io::Result<BenchmarkResult> {
-    println!(
-        "{}{}: {}",
-        "Benchmark #".bold(),
-        (num + 1).to_string().bold(),
-        cmd
-    );
+    if options.output_style != OutputStyleOption::None { 
+        println!(
+            "{}{}: {}",
+            "Benchmark #".bold(),
+            (num + 1).to_string().bold(),
+            cmd
+        );
+    }
 
     let mut times_real: Vec<Second> = vec![];
     let mut times_user: Vec<Second> = vec![];
@@ -207,11 +214,14 @@ pub fn run_benchmark(
 
     // Warmup phase
     if options.warmup_count > 0 {
-        let progress_bar = get_progress_bar(
-            options.warmup_count,
-            "Performing warmup runs",
-            options.output_style,
-        );
+        let mut progress_bar: Option<ProgressBar> = None;
+        if options.output_style != OutputStyleOption::None { 
+                progress_bar = Some(get_progress_bar(
+                options.warmup_count,
+                "Performing warmup runs",
+                options.output_style,
+            ));
+        }
 
         for _ in 0..options.warmup_count {
             let _ = time_shell_command(
@@ -221,17 +231,21 @@ pub fn run_benchmark(
                 options.failure_action,
                 None,
             )?;
-            progress_bar.inc(1);
+            if let Some(pbar) = progress_bar.clone() { pbar.inc(1); }
         }
-        progress_bar.finish_and_clear();
+        if let Some(bar) = progress_bar { bar.finish_and_clear(); }
     }
 
     // Set up progress bar (and spinner for initial measurement)
-    let progress_bar = get_progress_bar(
-        options.runs.min,
-        "Initial time measurement",
-        options.output_style,
-    );
+    let mut progress_bar:Option<ProgressBar> = None;
+
+    if options.output_style != OutputStyleOption::None { 
+            progress_bar = Some(get_progress_bar(
+            options.runs.min,
+            "Initial time measurement",
+            options.output_style,
+        ))
+    }
 
     // Run init command
     let prepare_cmd = options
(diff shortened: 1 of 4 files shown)


src/hyperfine/benchmark.rs
```diff
@@ -132,10 +137,10 @@ pub fn mean_shell_spawning_time(
                 times_user.push(r.time_user);
                 times_system.push(r.time_system);
             }
-        }
-        progress_bar.inc(1);
+        }  
+        // if let Some(bar) = progress_bar { bar.inc(1); }
```
sharkdp: what's going on here?

src/hyperfine/benchmark.rs
```diff
@@ -207,11 +214,14 @@ pub fn run_benchmark(
...
-            options.output_style,
-        );
+        let mut progress_bar: Option<ProgressBar> = None;
+        if options.output_style != OutputStyleOption::None { 
+                progress_bar = Some(get_progress_bar(
+                options.warmup_count,
+                "Performing warmup runs",
+                options.output_style,
+            ));
+        }
```
sharkdp: We could avoid `mut` here by writing this as
```
let progress_bar = if … {
  …
} else {
  …
};
```

src/hyperfine/benchmark.rs
```diff
@@ -221,17 +231,21 @@ pub fn run_benchmark(
                 options.failure_action,
                 None,
             )?;
-            progress_bar.inc(1);
+            if let Some(pbar) = progress_bar.clone() { pbar.inc(1); }
```
sharkdp: We should avoid `.clone()`ing. You can probably use `if let Some(ref pbar) = …`

src/hyperfine/benchmark.rs
```diff
@@ -221,17 +231,21 @@ pub fn run_benchmark(
...
-    );
+    let mut progress_bar:Option<ProgressBar> = None;
+
+    if options.output_style != OutputStyleOption::None { 
+            progress_bar = Some(get_progress_bar(
+            options.runs.min,
+            "Initial time measurement",
+            options.output_style,
+        ))
+    }
```
sharkdp: Same as above. can be constructed in a a single statement.

src/hyperfine/benchmark.rs
```diff
@@ -278,8 +292,8 @@ pub fn run_benchmark(
     all_succeeded = all_succeeded && success;
 
     // Re-configure the progress bar
-    progress_bar.set_length(count);
-    progress_bar.inc(1);
+    if let Some(bar) = progress_bar.clone() { bar.set_length(count); }
+    if let Some(bar) = progress_bar.clone() { bar.inc(1); }
```
sharkdp: same here. avoid the `.clone()`

Also. why not use a single `if`?

src/hyperfine/benchmark.rs
```diff
@@ -289,7 +303,8 @@ pub fn run_benchmark(
             let mean = format_duration(mean(&times_real), options.time_unit);
             format!("Current estimate: {}", mean.to_string().green())
         };
-        progress_bar.set_message(&msg);
+
+        if let Some(bar) = progress_bar.clone() { bar.set_message(&msg); }
```
sharkdp: same here. avoid the `.clone()`

src/hyperfine/benchmark.rs
```diff
@@ -305,9 +320,9 @@ pub fn run_benchmark(
 
         all_succeeded = all_succeeded && success;
 
-        progress_bar.inc(1);
+        if let Some(bar) = progress_bar.clone() { bar.inc(1); }
```
sharkdp: same here. avoid the `.clone()`

src/hyperfine/benchmark.rs
```diff
@@ -305,9 +320,9 @@ pub fn run_benchmark(
...
     }
-    progress_bar.finish_and_clear();
+    if let Some(bar) = progress_bar { bar.finish_and_clear(); }
```
sharkdp: same here. avoid the `.clone()`

src/hyperfine/benchmark.rs
```diff
@@ -370,15 +387,15 @@ pub fn run_benchmark(
         warnings.push(Warnings::OutliersDetected);
     }
 
-    if !warnings.is_empty() {
+    if !warnings.is_empty() && options.output_style != OutputStyleOption::None {
```
sharkdp: I think we should always print the warnings. They are written to STDERR anyway.

src/hyperfine/types.rs
```diff
@@ -113,6 +113,9 @@ pub enum OutputStyleOption {
 
     /// Keep coloring, but use no progress bar
     Color,
+
+    /// Disable all the output
+    None,
```
sharkdp: `None` is somehow associated with `Option<…>`. Maybe rename this to `Disabled`?

src/main.rs
```diff
@@ -54,8 +54,9 @@ fn main() {
 
     match res {
         Ok(timing_results) => {
-            write_benchmark_comparison(&timing_results);
-            let ans = export_manager.write_results(timing_results, options.unwrap().time_unit);
+            let unwrapped = options.unwrap();
```
sharkdp: instead of calling it `unwrapped`, we could call it `options` and shadow the original name.

The fact that we call `unwrap` here anyway is not really great, but that should be fixed elsewhere.

knidarkness:
Thanks for the review! Sorry for taking your time for those quite simple things, not too much in Rust yet and decided to improve myself in it during Hacktoberfest.

Will update up to your comments today/tomorrow.

knidarkness:
@sharkdp , I have updated the PR as you have described. Again, thanks a lot for lots of useful comments.

src/hyperfine/benchmark.rs
knidarkness: oh, that was just a bug due to my struggling with borrowing. fixed in the update

src/hyperfine/benchmark.rs
knidarkness: thanks for great suggestion!

src/hyperfine/benchmark.rs
knidarkness: Yeah, thanks, you were right, that works well and improves performance, although that (performance) doesn't seem crucial for this case?

src/hyperfine/benchmark.rs
knidarkness: updated

src/hyperfine/benchmark.rs
knidarkness: updated

hmm, yeah, good point about single `if`

src/hyperfine/benchmark.rs
knidarkness: fixed

src/hyperfine/benchmark.rs
knidarkness: fixed

src/hyperfine/types.rs
knidarkness: done, good point

src/main.rs
knidarkness: updated with shodowing for now. do you have any ideas on re-writing in the way not to have to `unwrap` the `options` object?

src/hyperfine/benchmark.rs
knidarkness: Agree on that point, changed in code

src/hyperfine/benchmark.rs
sharkdp: no, it's not. but I think its much cleaner without cloning it everywhere.

knidarkness: yes, totally agree on that

sharkdp: Coming to think of it, there is an even cleaner solution to write this, that we should probably follow. You can use `map` to "directly" call a function on the object inside the `Option`, allowing us to get rid of all the `if` statements. Should be something like
``` rust
progress_bar.map(|bar| bar.inc(1));
```
or
``` rust
progress_bar.as_ref().map(|bar| bar.inc(1));
```

knidarkness: Hmm, that looks quite interesting, but do you think it's that readable?

sharkdp: Yes, I think that's readable. Using `.map` (on `Option`, `Result`, or any kind of `Iterator`) is very common in Rust code.

knidarkness: Actually, seems quite nice when I researched a bit more. Updated the PR with that way

PR #228, opened 2019-10-14
Also run prepare command during warmup

### LINKED ISSUE #182: Should --prepare and --warmup be allowed at the same time?
If so (probably yes), we should also run the preparation command before each warmup run.

sseemayer:
Fixes #182

This was truly a "good first issue" to the point of it being that easy that I worry whether I misunderstood what the issue is about - please let me know if you need any further changes on this :smile:


sharkdp:
I think that's all, yes :smile:

Thank you very much!

PR #234, opened 2019-10-17
Add --parameter-list option

### LINKED ISSUE #227: Add new --parameter-list option for non-numeric parametrized benchmarks
Add a new option
```
--parameter-list <VAR> <VALUES>
```

that can be used like this:
```bash
hyperfine \
  --parameter-list branch_name master,feature1,bugfix,my-test-feature \
  --prepare "git checkout {branch_name}; make clean" \
  "make"
```

This would work just like the `--parameter-scan` option, but with non-numeric parameter values.

Maybe there is a better name for this command-line option.

see #216

JuanPotato:
This currently only splits the `VALUES` string on commas and doesn't try to do anything to support elements with commas in them. I couldn't find anything in clap that would support automatically parsing a list.

Closes #227

### DIFF
--- a/src/hyperfine/app.rs
+++ b/src/hyperfine/app.rs
@@ -140,8 +140,8 @@ fn build_app() -> App<'static, 'static> {
                 .help(
                     "Perform benchmark runs for each value in the comma-seperated list VALUES. \
                      Replaces the string '{VAR}' in each command by the current parameter value\
-                     .\n\nExample:  hyperfine -L threads 1,2,4 'make -j {threads}'\n\n\
-                     This performs benchmarks for 'make -j 1', 'make -j 2', and 'make -j 4'.",
+                     .\n\nExample:  hyperfine -L compiler gcc,clang '{compiler} -O2 main.cpp'\n\n\
+                     This performs benchmarks for 'gcc -O2 main.cpp' and 'clang -O2 main.cpp'.",
                 ),
         )
         .arg(
--- a/src/hyperfine/app.rs
+++ b/src/hyperfine/app.rs
@@ -136,7 +136,7 @@ fn build_app() -> App<'static, 'static> {
                 .takes_value(true)
                 .allow_hyphen_values(true)
                 .value_names(&["VAR", "VALUES"])
-                .conflicts_with("parameter-scan")
+                .conflicts_with_all(&["parameter-scan", "parameter-step-size"])
                 .help(
                     "Perform benchmark runs for each value in the comma-seperated list VALUES. \
                      Replaces the string '{VAR}' in each command by the current parameter value\
--- a/src/hyperfine/app.rs
+++ b/src/hyperfine/app.rs
@@ -130,6 +130,21 @@ fn build_app() -> App<'static, 'static> {
                 ),
         )
         .arg(
+            Arg::with_name("parameter-list")
+                .long("parameter-list")
+                .short("L")
+                .takes_value(true)
+                .allow_hyphen_values(true)
+                .value_names(&["VAR", "VALUES"])
+                .conflicts_with("parameter-scan")
+                .help(
+                    "Perform benchmark runs for each value in the comma-seperated list VALUES. \
+                     Replaces the string '{VAR}' in each command by the current parameter value\
+                     .\n\nExample:  hyperfine -L threads 1,2,4 'make -j {threads}'\n\n\
+                     This performs benchmarks for 'make -j 1', 'make -j 2', and 'make -j 4'.",
+                ),
+        )
+        .arg(
             Arg::with_name("style")
                 .long("style")
                 .short("s")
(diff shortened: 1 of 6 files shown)


src/hyperfine/app.rs
```diff
@@ -129,6 +129,21 @@ fn build_app() -> App<'static, 'static> {
...
+                .long("parameter-list")
+                .short("L")
+                .takes_value(true)
+                .allow_hyphen_values(true)
+                .value_names(&["VAR", "VALUES"])
+                .conflicts_with("parameter-scan")
+                .help(
+                    "Perform benchmark runs for each value in the comma-seperated list VALUES. \
+                     Replaces the string '{VAR}' in each command by the current parameter value\
+                     .\n\nExample:  hyperfine -L threads 1,2,4 'make -j {threads}'\n\n\
```
sharkdp: Can we find a simple example with non-numeric parameter values?

JuanPotato: Perhaps `hyperfine -L file a,b,c "gcc {file}.cpp"` ?

src/hyperfine/types.rs
```diff
@@ -60,7 +60,7 @@ impl<'a> Command<'a> {
     pub fn new_parametrized(
         expression: &'a str,
         parameter: &'a str,
-        value: NumericType,
+        value: String,
```
sharkdp: If you take a `&str` here instead of a `String`, you can avoid many of the `.clone` calls.

src/hyperfine/types.rs
```diff
@@ -46,7 +46,7 @@ pub struct Command<'a> {
     expression: &'a str,
 
     /// A possible parameter value.
-    parameter: Option<(&'a str, NumericType)>,
+    parameter: Option<(&'a str, String)>,
```
sharkdp: It might even be possible to use a `&'a str` reference here instead of a `String`.

src/hyperfine/types.rs
```diff
@@ -78,8 +78,8 @@ impl<'a> Command<'a> {
         }
     }
 
-    pub fn get_parameter(&self) -> Option<(&'a str, NumericType)> {
-        self.parameter
+    pub fn get_parameter(&self) -> &Option<(&'a str, String)> {
```
sharkdp: similar here, this could probably return a `&'a str` ref, instead of a `String`

src/hyperfine/app.rs
```diff
@@ -129,6 +129,21 @@ fn build_app() -> App<'static, 'static> {
...
+        .arg(
+            Arg::with_name("parameter-list")
+                .long("parameter-list")
+                .short("L")
+                .takes_value(true)
+                .allow_hyphen_values(true)
+                .value_names(&["VAR", "VALUES"])
+                .conflicts_with("parameter-scan")
+                .help(
+                    "Perform benchmark runs for each value in the comma-seperated list VALUES. \
```
sharkdp: What do you think: we could rename `VALUES` to `LIST`. This way, we could write: *"Perform benchmark runs for each value in the comma-seperated LIST"*. Not sure if that's necessarily better/clearer.

src/hyperfine/types.rs
JuanPotato: I think the reason I did String was because we call `new_parameterized` for the parameter range option, we passed a `NumericType` which implements Clone. If we wanted to pass `&str` we would have to turn the `NumericType` into a `String` first and then take a reference of that, but if you look at `build_parameterized_commands`, where we call `new_parameterized`, it doesn't look like there is a very good way of doing that.

src/hyperfine/app.rs
sharkdp: how about:
```c++
hyperfine -L compiler gcc,clang "{compiler} -O3 -Wall main.c"
```

sharkdp: That's actually really useful when I compare different versions of my programs, because I will not have to repeat the arguments over and over again. For example:
``` bash
hyperfine -L version 6.0,7.1,7.2,7.4,master "./fd-{version} -HI 'my-pattern' /some/directory"
```

JuanPotato:
The parameter option seems to export the parameter just fine. Only difference is that now the parameter value is enclosed in quotes since it is a string.

src/hyperfine/app.rs
JuanPotato: I like the different program versions a lot

sharkdp:
I added a commit which changes the way that a parameter value is stored internally (as an `enum` rather than a `String`). Hope that's fine with you.

sharkdp:
> The parameter option seems to export the parameter just fine. Only difference is that now the parameter value is enclosed in quotes since it is a string.

The old parameter value was also exported as a string, so that's fine.

sharkdp:
I also added a simple tokenizer that allows us to escape commas:
```bash
hyperfine -L my_parameter 'hello\, world!,other value' …
```

sharkdp:
Thank you!

PR #250, opened 2019-12-22
Reflecting inclusion in Ubuntu Universe.

0xf2:
I suggest changing install to apt if there are no known issues with the Universe package.


sharkdp:
I... don't think there is an official package. Or is there?

For Debian, there is an ITP-request ("intent to package"): https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=908776

Issue #212, opened 2019-09-16
Man page?


Roguelazer:
I don't suppose anyone has a man page floating around? It'd be great to have as part of packaging this tool...

sharkdp:
No, I don't think anyone wrote a man page. If somebody wants to work on this, I'm happy to include it in `hyperfine` packages.

```
help2man -N -n hyperfine hyperfine > hyperfine.1
```
can serve as a good start.

sharkdp:
Closed via #257 by @cadeef

sharkdp:
Released in [v1.10.0](https://github.com/sharkdp/hyperfine/releases/tag/v1.10.0).

PR #257, opened 2020-01-11
Add basic manpage

cadeef:
Addresses #212 
This manpage is loosely converted from `help2man` and then modified to convention.

### DIFF
--- a/doc/hyperfine.1
+++ b/doc/hyperfine.1
@@ -173,10 +173,10 @@ Print help message.
 Show version information.
 .SH EXAMPLES
 .LP
-Basic benchmark of 'find . -iname '*[0-9].jpg'':
+Basic benchmark of 'find . -name todo.txt':
 .RS
 .nf
-\fBhyperfine\fR 'find . -iname '*[0-9].jpg''
+\fBhyperfine\fR 'find . -name todo.txt'
 .fi
 .RE
 .LP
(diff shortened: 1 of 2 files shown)


doc/hyperfine.1
```diff
@@ -0,0 +1,212 @@
...
+\fB\-h\fR, \fB\-\-help\fR
+.IP
+Print help message.
+.HP
+\fB\-V\fR, \fB\-\-version\fR
+.IP
+Show version information.
+.SH EXAMPLES
+.LP
+Basic benchmark of 'find . -iname '*[0-9].jpg'':
```
sharkdp: the nested quotes look a bit weird here (and would yield an error in the command line below). Maybe just:
```
Basic benchmark of 'find . -name todo.txt'
```
and then:
```roff
\fBhyperfine\fR 'find . -name todo.txt' 
```

doc/hyperfine.1
```diff
@@ -0,0 +1,212 @@
...
+.LP
+Perform benchmarks for 'sleep 0.2' and 'sleep 3.2' with a minimum 5 runs each:
+.RS
+.nf
+\fBhyperfine\fR \fB\-\-min\-runs\fR 5 'sleep 0.2' 'sleep 3.2'
+.fi
+.RE
+.LP
+Perform a benchmark of 'grep' with a warm disk cache by first running the command
+3 times:
```
sharkdp: maybe: "… with a warm disk cache by performing 3 runs upfront that are not part of the measurement"?

doc/hyperfine.1
cadeef: Heh, I added that one after I verified the commands and missed it in the final review.

PR #274, opened 2020-03-20
Update plot_histogram.py

ghaiklor:
It is more natural to see time by Y axis

Figure_1


sharkdp:



> It is more natural to see time by Y axis

I disagree. `plot_histogram.py` basically shows a probability distribution over runtime. If you have enough benchmark runs, this is how it could look like:

image

I think it's more natural to show probability/counts over time, instead of the other way around.

ghaiklor:
Makes sense when you explained

PR #281, opened 2020-04-01
README: add instructions on installing via MacPorts for macOS

herbygillot:
First off, thank you for `hyperfine` - it's extremely useful and a very nice utility.
It's been added to [MacPorts](https://www.macports.org), so just adding instructions on installing via that method.


sharkdp:
Thank you very much for your feedback and for your contribution!

Issue #276, opened 2020-03-31
Feature request: Option to use as a simple command timer?


michaelmoussa:
My first thought upon discovering this great utility was to do `alias time="hyperfine --runs 1"` and use it whenever I'd normally use the `time` command as well. That, understandably, isn't allowed due to needing two or more runs to compute variance, but I still want to do it! :)

Would you merge a PR that adds something along the lines of the following?

```bash
    -t, --time
        Perform a single run and display its timing, omitting any
        extended statistics. Implies "--runs 1" and "--show-output"
```

I realize this is probably an edge case and _not_ the primary intent of this utility, but I figured I'd ask since it doesn't seem like too much of a stretch. Regardless, thanks for your work on this!

sharkdp:
Thank you very much for your feedback!

That sounds like a reasonable request, but what exactly would be the benefit over using `time`?

We would have to omit almost every field from the output:
![image](https://user-images.githubusercontent.com/4209276/78038299-fad2c280-736c-11ea-8744-b6de4b190ac1.png)

* All of "sigma", "min", "max" and "# runs" would be useless
* I guess we could still call it "mean", but it would be misleading
* "User" and "System" are still fine.

michaelmoussa:
The benefits are admittedly *very* minor, so hopefully the level of effort to implement would be the same.

* Millisecond display by default vs. `time` that always uses seconds.
* Colorized output of `hyperfine` is much easier to read at a glance. If I needed to, I'd probably `alias time="hyperfine --runs 2"` just for that.
* I use `time` for individual time checks much more often than I do any kind of benchmarking. Being able to use hyperfine for individual command timing as well guarantees that I'll remember to use it the next time I need to benchmark, versus scrolling through my GitHub stars thinking "hmm... what was that cool command benchmark tool I saw recently called again?"

re: omitting fields, I don't think that'd be strictly necessary. The default behavior of automatically determining a number of runs to use would be unchanged. Explicitly using the `--time` flag means the user is intentionally limiting it to only one run. Yes, a single run *does* make the `min`, `max`, `# runs`, and `mean` values useless, it *doesn't* make them objectively wrong, and the person using it shouldn't be surprised that `min = max = mean` and that the variance is 0.

sharkdp:
Ok, if we can leave the default layout the same and just output the values as is (even if meaningless), I'm fine with this change.

We could even think about hiding this feature from the `--help` text (it could still be documented in the man page) in order not to confuse the "average user".

[...]

PR #282, opened 2020-04-06
Option to use as a simple command timer

michaelmoussa:
Initial attempt at implementing #276. This is my first attempt at working with Rust, so pardon any beginner mistakes. :)

Rather than adding a "hidden" `-t, --time` option, I simply made it possible to set `--runs=1` by setting a default standard deviation of `0.0` when doing single runs. The error from the statistical library is to prevent the sample standard deviation formula's `N-1` denominator from attempting a divide by zero (which I assume is why `runs` originally had to be `>=2`). This makes sense when benchmarking, as we are not sampling the *entire* population. In the case of `--runs=1`, we're just measuring the time it took the command to execute once. We aren't trying to infer anything about the population, and a standard deviation of `0.0` when all data points (in this case, just one of them) have the same value is valid.

**Example usage below:**

image

As for the following comment regarding mixing `--show-output` and `--style`...

> "I think that's actually overly restrictive. What we originally wanted (if I remember correctly) was to disable the interactive components (the progress bar and spinner). Colorized output (`--style=color` instead of `--style=full`) should be fine. Maybe we can change that."

I changed it so that `--style` and `--show-output` no longer conflict, but I wasn't clear if any other special handling was needed. Is this fine, or should `--show-output` still conflict with `--style=full`?

### DIFF
--- a/src/hyperfine/app.rs
+++ b/src/hyperfine/app.rs
@@ -216,7 +216,6 @@ fn build_app() -> App<'static, 'static> {
         .arg(
             Arg::with_name("show-output")
                 .long("show-output")
-                .conflicts_with("style")
                 .help(
                     "Print the stdout and stderr of the benchmark instead of suppressing it. \
                      This will increase the time it takes for benchmarks to run, \
--- a/src/hyperfine/benchmark.rs
+++ b/src/hyperfine/benchmark.rs
@@ -342,7 +342,11 @@ pub fn run_benchmark(
     // Compute statistical quantities
     let t_num = times_real.len();
     let t_mean = mean(&times_real);
-    let t_stddev = standard_deviation(&times_real, Some(t_mean));
+    let t_stddev = if times_real.len() > 1 {
+        standard_deviation(&times_real, Some(t_mean))
+    } else {
+        0.0
+    };
     let t_median = median(&times_real);
     let t_min = min(&times_real);
     let t_max = max(&times_real);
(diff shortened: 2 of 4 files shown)


michaelmoussa:
@sharkdp just checking in on this. Is there anything else I need to do in order for it to be merged?

sharkdp:
> Rather than adding a "hidden" `-t, --time` option, I simply made it possible to set `--runs=1` by setting a default standard deviation of `0.0` when doing single runs.

I don't really like that. It prevents a potentially helpful error message if someone does this by accident (or via a script). The old error message should probably be improved though:

*Error: The number of runs must be two or higher in order to get an estimate for the standard deviation. Use the '--time' option if you want to allow single-run measurements.*

> (which I assume is why `runs` originally had to be `>=2`).

exactly.

> and a standard deviation of `0.0` when all data points (in this case, just one of them) have the same value is valid.

well. I'd rather say that you cannot infer anything about the standard deviation if you take just a single measurement. Printing `0.0` is potentially confusing. An actual benchmark with a huge amount of runs could show a (rounded) value of 0.0 when the standard deviation much much smaller than the measured time.

If we add a `--time` option, we could add a help text that explicitly mentions this (that the standard deviation will be 0.0).

sharkdp:
Closing this for now, unless there is further feedback.

sharkdp:
There is now a cleaner implementation of this in #447.

PR #308, opened 2020-07-29
Remove manpage comment

cadeef:
The comment causes a syntax error when run with older versions of man (i.e. 1.6g on MacOS)


sharkdp:
Thank you very much!

PR #313, opened 2020-08-10
Feature/incremental export

BaronVonDrew:
Adds the ability to write CSV results incrementally to a file, rather than waiting till all of the tests have finished to write export the results using the `incremental-export` option.

Makes the `CsvExporter::new` method require the path so it can maintain an open writer to write the output to a file. As the exporter contains a writer, and calls to `write_incremental_results` mutate the internal writer, this and the ExportManager need to be borrowed mutably by the `run()` function.

### DIFF
--- a/src/hyperfine/app.rs
+++ b/src/hyperfine/app.rs
@@ -229,10 +229,8 @@ fn build_app() -> App<'static, 'static> {
                 .long("incremental-export")
                 .requires("export-csv")
                 .help(
-                    "Print the stdout and stderr of the benchmark instead of suppressing it. \
-                     This will increase the time it takes for benchmarks to run, \
-                     so it should only be used for debugging purposes or \
-                     when trying to benchmark output speed.",
+                    "Writes test results incrementally to a file after each test run \
+                    instead of waiting till all tests have finished.",
                 ),
         )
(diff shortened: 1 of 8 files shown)


sharkdp:
Thank you very much for your contribution. I would really appreciate if we could find a way to do this without adding new command line options. Would there be any harm in changing the existing CSV export to be incremental? I/O happens between two benchmark runs, so I don't think it would affect any timings (or could it?).

BaronVonDrew:
No, I suspect you're right, there probably won't be much impact, I'll refactor this PR to write CSV incrementally by default.

BaronVonDrew:
Updated to remove the `--incremental-write` flag. Now, if the exporter supports incremental writing it will write incrementally.

sharkdp:
Thank you for the updates!

I thought about this again and had an idea how we could hopefully add incremental export for all formats, while keeping the code changes minimal.

Couldn't we simply run the existing export after each benchmarking run (instead of only at the end)? This would not really be "incremental", but the export should be so fast that it doesn't really matter whether or not we are re-writing the old results again, right?

sharkdp:
Closing this in favor of #339. @BaronVonDrew Thank you very much for the initial work here!

PR #314, opened 2020-08-27
Switch to owo-colors.

inspier:
This pull request switches out the [colored](https://github.com/mackwic/colored) crate for [owo-colors](https://github.com/jam1garner/owo-colors).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -97,17 +97,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "colored"
-version = "2.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b3616f750b84d8f0de8a58bda93e08e2a81ad3f523089b05f1dffecab48c6cbd"
-dependencies = [
- "atty",
- "lazy_static",
- "winapi",
-]
-
-[[package]]
 name = "console"
 version = "0.11.3"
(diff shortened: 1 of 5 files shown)


sharkdp:

Why would we want to switch?

inspier:
owo-colors has no dependencies and is no_std compatible, while the latter may not really be applicable to hyperfine it is perhaps still a win.

sharkdp:
colored has served us well so far. Also, it's not like colored is not maintained anymore, right?

I don't really see a reason to switch to another dependency if there is no need. If the dependency tree is significantly smaller, that might be a reason to switch though.

inspier:
`colored` has three dependencies atty, lazy_static, and winapi. While owo-colors has no dependencies and a near identical API for the most part. That said you make a fair point about not fixing something that isn't broken. So feel free to close this pull request or let me know if that is something I need to do.

sharkdp:
> `colored` has three dependencies atty, lazy_static, and winapi

all of these are dependencies of `hyperfine` anyway.

> So feel free to close this pull request

I will close it for now, but we can come back to it if the situation changes. Thank you very much for your contribution.

sharkdp:
Maybe the main reason not to introduce new dependencies is that package maintainers often have a hard time to package the new dependencies.

PR #318, opened 2020-09-15
Support matrices of independent parameter lists

### LINKED ISSUE #253: Feature Request: Matrix measurements
If I want to benchmark my program with one variable command argument, I can use `--parameter-list`. On the other hand, if I want to benchmark four, five, six variables, it can get tedious.

It would be nice if I'd be able to run:

```
hyperfine 'arg {a1} {a2} {a3}' --parameter-list a1 1,2,3 --parameter-list a2 a,b,c --parameter-list a3 Q,W,E
```

wchargin:
This patch permits the `-L`/`--parameter-list` argument to appear
multiple times. Each additional parameter increases the dimensionality
of the parameter matrix. One benchmark will be run for each combination
of parameters (i.e., the benchmark space is the Cartesian product of the
parameter spaces).

For now, `--parameter-list` and `--parameter-scan` are still mutually
exclusive. If desired, a follow-up change could similarly permit
multiple occurrences of `--parameter-scan` and also permit use of both
listed and scanned parameters together. (After all, `--parameter-scan`
can be thought of as syntactic sugar for a parameter list.)

Benchmark results now include a `parameters` dictionary mapping
parameter names to values (as before, omitted when not parameterized).
This means that we can’t use `serde` to serialize to CSV, but a CSV
writer is easy to write by hand, so we do that. This also means that the
JSON output format has changed: write `b.parameters["foo"]` rather than
`b.parameter`.

The `plot_parametrized.py` helper script still requires that there be
exactly one parameter among the input files. We could allow users to
specify one parameter to marginalize over, but then we’d need to
aggregate over the other parameters, which is a bit out of the scope of
this change. The `--parameter-name` argument has been removed, since it
can now be inferred from the output dictionary.

This implementation is a little profligate with memory usage for
`BenchmarkResult`s: each result contains a separate copy of all
parameter names (as `String`s). This could be done away with by
threading lifetimes through the `BenchmarkResult`s, or by ref-counting
the names.

Fixes #253.

Test Plan:
Some unit tests included. As an integration test, try:

```
cargo run --release -- --runs 10 \
    -L foo foo1,foo2 -L bar bar9,bar8 \
    'echo {foo} {bar}' \
    'printf "%s\n" {foo} {bar}' \
    --export-csv /tmp/out
```

with all the export formats.

Screenshot of new `plot_parametrized.py` with two input files, no
`--parameter-name` argument:

```
$ hyperfine -P my_favorite_parameter 1 4 'echo {my_favorite_parameter}' -w 3 --export-json /tmp/echo
$ hyperfine -P my_favorite_parameter 1 4 'printf {my_favorite_parameter}' -w 3 --export-json /tmp/printf
$ python ./scripts/plot_parametrized.py /tmp/echo /tmp/printf
```

![Screenshot showing an error bar chart with two series, with an x-axis
label of “my_favorite_parameter”][plot-ss]

[plot-ss]: https://user-images.githubusercontent.com/4317806/93403577-c5e3fa80-f83c-11ea-8bae-982ac3136040.png

wchargin-branch: param-list-matrix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,23 @@
 # unreleased
 
 ## Features
+
+- The `-L`/`--parameter-list` option can now be specified multiple times to evaluate all possible combinations of the listed parameters:
+
+  ``` bash
+  hyperfine -L number 1,2 -L letter a,b,c \
+      "echo {number}{letter}" \
+      "printf '%s\n' {number}{letter}"
+  # runs 12 benchmarks: 2 commands (echo and printf) times 6 combinations of
+  # the "letter" and "number" parameters
+  ```
+
+  See: #253, #318 (@wchargin)
+
 ## Changes
+
+- When parameters are used with `--parameter-list` or `--parameter-scan`, the JSON export format now contains a dictionary `parameters` instead of a single key `parameter`.
+
 ## Bugfixes
 ## Other
 ## Packaging
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,8 @@
 
 ## Changes
 
-- When parameters are used with `--parameter-list` or `--parameter-scan`, the JSON export format now contains a dictionary `parameters` instead of a single key `parameter`.
+- When parameters are used with `--parameter-list` or `--parameter-scan`, the JSON export format now contains a dictionary `parameters` instead of a single key `parameter`. See #253, #318.
+- The `plot_parametrized.py` script now infers the parameter name, and its `--parameter-name` argument has been deprecated. See #253, #318.
 
 ## Bugfixes
 ## Other
--- a/src/hyperfine/types.rs
+++ b/src/hyperfine/types.rs
@@ -1,6 +1,7 @@
 use rust_decimal::Decimal;
 /// This module contains common internal types.
 use serde::*;
+use std::collections::BTreeMap;
 use std::fmt;
 
 use crate::hyperfine::units::{Second, Unit};
@@ -11,7 +12,7 @@ pub const DEFAULT_SHELL: &str = "sh";
 #[cfg(windows)]
 pub const DEFAULT_SHELL: &str = "cmd.exe";
 
-#[derive(Debug, Clone, Serialize, Copy, PartialEq)]
+#[derive(Debug, Clone, Serialize, Copy, PartialEq, Eq)]
 #[serde(untagged)]
 pub enum NumericType {
     Int(i32),
@@ -39,7 +40,7 @@ impl Into<NumericType> for Decimal {
     }
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum ParameterValue {
     Text(String),
     Numeric(NumericType),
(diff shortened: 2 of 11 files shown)


wchargin:
I should note that this changes the format of the `--export-json` format
by replacing the `parameter` key with a `parameters` object. If that’s
considered a breaking change, feel free to take it into account.

There are also a few ways in which this implementation could be made
more efficient: by computing the command templates just once per command
rather than for each set of parameter values (and indeed for each call
to `get_shell_command`), or by build the set of commands lazily as an
`impl Iterator<Item = Command<'a>>` rather than a `Vec<Command<'a>>`. In
practice, I don’t expect this to be a big deal—in my experience, the
slow bits are always in the programs being benchmarked, not `hyperfine`
itself—so I’ve left them as is for now to avoid scope creep.

albins:
Nice! I'll have a look and see how this works on my use case; I've really been looking for something like this!

sharkdp:

This looks great. I hope to find time for a proper review soon.

A few comments upfront:

> I should note that this changes the format of the `--export-json` output
> by replacing the `parameter` key with a `parameters` object. If that’s
> considered a breaking change, feel free to take it into account.

The JSON output is the only format where I would indeed care about backwards compatibility, because it's mainly intended for interoperability with other tools. The Python tools in `scripts/` use the JSON format to post-process `hyperfine` benchmark results. We should make sure that they can handle the new format (I would guess that `plot_parametrized.py` is the only affected script. We do not have to support matrix benchmarks here, but we should maybe throw a nice error message..).

Apart from that, I am fine with the breaking change here - especially if it allows for a new feature like this. I like that we now also include the parameter name in the JSON format. Ideally, we would make use of this in the `plot_parametrized.py` script and remove the `--parameter-name` option.

> There are also a few ways in which this implementation could be made
> more efficient […]

> I don’t expect this to be a big deal—in my experience, the
> slow bits are always in the programs being benchmarked, not `hyperfine`
> itself

Absolutely. Let's focus on code clarity instead.

CHANGELOG.md
```diff
@@ -1,7 +1,23 @@
...
+      "printf '%s\n' {number}{letter}"
+  # runs 12 benchmarks: 2 commands (echo and printf) times 6 combinations of
+  # the "letter" and "number" parameters
+  ```
+
+  See: #253, #318 (@wchargin)
+
 ## Changes
+
+- When parameters are used with `--parameter-list` or `--parameter-scan`, the JSON export format now contains a dictionary `parameters` instead of a single key `parameter`.
```
sharkdp, line 19: :+1:

src/hyperfine/types.rs
```diff
@@ -55,49 +56,79 @@ impl<'a> ToString for ParameterValue {
...
+                format!("{{{param_name}}}", param_name = param_name),
+                param_value.to_string(),
+            );
         }
+        let mut remaining = self.expression;
+        // Manually replace consecutive occurrences to avoid double-replacing: e.g.,
+        //
+        //     hyperfine -L foo 'a,{bar}' -L bar 'baz,quux' 'echo {foo} {bar}'
+        //
+        // should not ever run 'echo baz baz'. See `test_get_shell_command_nonoverlapping`.
```
sharkdp, line 100: Very nice.

wchargin:
Thanks for the comments! Updated `plot_parametrized.py` to infer the
parameter name. I’ve left the `--parameter-name` option in to not
immediately break callers, but it’s marked as deprecated and does
nothing other than print a warning. I also checked that the other
scripts still work without modification.

I’ve only tested the Python code on Python 3, since Python 2 is dead.

Happy to incorporate any more feedback that you may have.

sharkdp (approved):
Fantastic work! I have nothing else to add.

Thank you very much for working on this and for contributing such a high quality PR!

Sarcasm:
Great!

> For now, --parameter-list and --parameter-scan are still mutually
> exclusive. If desired, a follow-up change could similarly permit
> multiple occurrences of --parameter-scan and also permit use of both
> listed and scanned parameters together. (After all, --parameter-scan
> can be thought of as syntactic sugar for a parameter list.)

Personally I think this follow up PR would make sense, that is what I needed at some point.
I wanted to test a parameter-scan (number of threads), against a parameter-list (a list of filesystems).

PR #321, opened 2020-09-28
Add Maximum Resident Set tracking

### LINKED ISSUE #86: Support for benchmarking memory usage, etc.
This is definitely not in scope is hyperfine's goal is to be just a replacement for `time`, but the thing I most wish for when using time to benchmark programs is if it could measure other relevant program execution statistics, like memory usage. After all, optimization (which is what benchmarking is driving) is always about tradeoffs in some way. But it's kinda hard to know if you're making the right tradeoffs for your project if you're measuring only one thing (here, execution speed).

Would something measuring memory usage, etc. that be in scope for hyperfine a a benchmarking tool?

lu-zero:
Fixes #86

### DIFF
--- a/src/hyperfine/benchmark.rs
+++ b/src/hyperfine/benchmark.rs
@@ -5,7 +5,9 @@ use std::process::Stdio;
 use colored::*;
 use statistical::{mean, median, standard_deviation};
 
-use crate::hyperfine::format::{format_duration, format_duration_unit};
+use crate::hyperfine::format::{
+    format_duration, format_duration_unit, format_memory, format_memory_unit,
+};
 use crate::hyperfine::internal::{get_progress_bar, max, min, MIN_EXECUTION_TIME};
 use crate::hyperfine::outlier_detection::{modified_zscores, OUTLIER_THRESHOLD};
 use crate::hyperfine::shell::execute_and_time;
@@ -14,7 +16,7 @@ use crate::hyperfine::timer::{TimerStart, TimerStop};
 use crate::hyperfine::types::{
     BenchmarkResult, CmdFailureAction, Command, HyperfineOptions, OutputStyleOption,
 };
-use crate::hyperfine::units::Second;
+use crate::hyperfine::units::{MebiByte, Second};
 use crate::hyperfine::warnings::Warnings;
 
 /// Results from timing a single shell command
@@ -28,14 +30,20 @@ pub struct TimingResult {
 
     /// Time spent in kernel mode
     pub time_system: Second,
+
+    /// Maximum resident set
+    pub max_rss: MebiByte,
 }
 
 /// Correct for shell spawning time
-fn subtract_shell_spawning_time(time: Second, shell_spawning_time: Second) -> Second {
-    if time < shell_spawning_time {
-        0.0
+fn saturating_subtract<T: std::ops::Sub<T> + PartialOrd<T> + Copy>(
+    resource: T,
+    shell_spawining_resource: T,
+) -> <T as std::ops::Sub>::Output {
+    if resource < shell_spawining_resource {
+        resource - resource
     } else {
-        time - shell_spawning_time
+        resource - shell_spawining_resource
     }
 }
 
@@ -59,6 +67,7 @@ pub fn time_shell_command(
 
     let mut time_user = result.user_time;
     let mut time_system = result.system_time;
+    let mut max_rss = result.max_rss;
 
     if failure_action == CmdFailureAction::RaiseError && !result.status.success() {
         return Err(io::Error::new(
@@ -71,9 +80,10 @@ pub fn time_shell_command(
 
     // Correct for shell spawning time
     if let Some(spawning_time) = shell_spawning_time {
-        time_real = subtract_shell_spawning_time(time_real, spawning_time.time_real);
-        time_user = subtract_shell_spawning_time(time_user, spawning_time.time_user);
-        time_system = subtract_shell_spawning_time(time_system, spawning_time.time_system);
+        time_real = saturating_subtract(time_real, spawning_time.time_real);
+        time_user = saturating_subtract(time_user, spawning_time.time_user);
+        time_system = saturating_subtract(time_system, spawning_time.time_system);
+        max_rss = saturating_subtract(max_rss, spawning_time.max_rss);
     }
 
     Ok((
@@ -81,6 +91,7 @@ pub fn time_shell_command(
             time_real,
             time_user,
             time_system,
+            max_rss,
         },
         result.status.success(),
     ))
@@ -106,6 +117,7 @@ pub fn mean_shell_spawning_time(
     let mut times_real: Vec<Second> = vec![];
     let mut times_user: Vec<Second> = vec![];
     let mut times_system: Vec<Second> = vec![];
+    let mut max_rss: Vec<MebiByte> = vec![];
 
     for _ in 0..COUNT {
         // Just run the shell without any command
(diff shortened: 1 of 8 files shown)


lu-zero:
It isn't complete but should work well as proof of concept

sharkdp:
Thank you very much for your contribution! I am planning to take a closer look soon.

src/hyperfine/benchmark.rs
```diff
@@ -74,13 +78,15 @@ pub fn time_shell_command(
         time_real = subtract_shell_spawning_time(time_real, spawning_time.time_real);
         time_user = subtract_shell_spawning_time(time_user, spawning_time.time_user);
         time_system = subtract_shell_spawning_time(time_system, spawning_time.time_system);
+        max_rss = subtract_shell_spawning_time(max_rss, spawning_time.max_rss);
```
sharkdp, line 81: Okay, so this subtracts the memory required by the shell? In this case, we should probably give the function a new name?

src/hyperfine/timer/unix_timer.rs
```diff
@@ -55,6 +55,7 @@ fn get_cpu_times() -> CPUTimes {
             + i64::from(result.ru_utime.tv_usec),
         system_usec: i64::from(result.ru_stime.tv_sec) * MICROSEC_PER_SEC
             + i64::from(result.ru_stime.tv_usec),
+        max_rss_byte: i64::from(result.ru_maxrss),
```
sharkdp, line 58: From `man 2 getrusage`: "in kilo-bytes"
```
       ru_maxrss (since Linux 2.6.32)
              This is the maximum resident  set  size  used  (in  kilo‐
              bytes).   For  RUSAGE_CHILDREN,  this is the resident set
              size of the largest child, not the maximum  resident  set
              size of the process tree.
```

src/hyperfine/timer/unix_timer.rs
```diff
@@ -63,6 +64,7 @@ fn cpu_time_interval(start: &CPUTimes, end: &CPUTimes) -> CPUInterval {
     CPUInterval {
         user: ((end.user_usec - start.user_usec) as f64) * 1e-6,
         system: ((end.system_usec - start.system_usec) as f64) * 1e-6,
+        max_rss: ((end.max_rss_byte - start.max_rss_byte) as f64) * 1e-6,
```
sharkdp, line 67: I would prefer if we used kibibytes and mebibytes here, i.e. divide by 1024 or 1024². All references of `MegaBytes` should then be changed to `KibiBytes` or `MebiBytes`. In the output we would ideally use "KiB" or "MiB".

This is to avoid the confusion with "KB" and "MB" where it's unclear if they refer to the SI standard (as they should, in my opinion), or if it's an old Unix program which uses "KB" and "MB" when they actually refer to "1024 byte" or "1024² byte".

src/hyperfine/timer/unix_timer.rs
sharkdp: Also, I'm not sure what to make of the "For  RUSAGE_CHILDREN,  this is the resident set size of the largest child, not the maximum  resident  set size of the process tree" comment. Does this mean that we would not get the total amount of memory used by the tree
```
+ /bin/sh
+--+ command-to-be-benchmarked
```
but rather the memory usage of EITHER the shell or the command-to-be-benchmarked?!

src/hyperfine/benchmark.rs
```diff
@@ -358,16 +371,18 @@ pub fn run_benchmark(
 
     let (user_str, user_unit) = format_duration_unit(user_mean, None);
     let system_str = format_duration(system_mean, Some(user_unit));
+    let max_rss_str = format!("{:.3} MB", max_rss_mean);
 
     if options.output_style != OutputStyleOption::Disabled {
         println!(
-            "  Time ({} ± {}):     {:>8} ± {:>8}    [User: {}, System: {}]",
+            "  Time ({} ± {}):     {:>8} ± {:>8}    [User: {}, System: {}, MaxRSS: {}]",
```
sharkdp, line 378: I'm pretty sure that most users would not know what "MaxRSS" refers to. Maybe just "Mem:" and then explain the precise meaning in the `--help` text and man page?

src/hyperfine/benchmark.rs
```diff
@@ -358,16 +371,18 @@ pub fn run_benchmark(
 
     let (user_str, user_unit) = format_duration_unit(user_mean, None);
     let system_str = format_duration(system_mean, Some(user_unit));
+    let max_rss_str = format!("{:.3} MB", max_rss_mean);
```
sharkdp, line 374: I would prefer if we would show either KiB or MiB (depending on the actual usage), with fewer digits in the fractional part.

sharkdp (changes requested):
Thank you for your contribution. I added a few comments. 

I am inclined to accept such a change, but only if
- it is thoroughly tested (not necessarily a unit test, but at least some manual tests on Linux and Windows)
- all questions here and in the ticket can be answered
- it nicely integrates within the code base (the namings of some structs/types/variables/functions are currently focused on times, not memory consumption)

src/hyperfine/benchmark.rs
lu-zero: Indeed, I did the least amount of changes to get something going, if you have idea on function names please tell me :)

src/hyperfine/timer/unix_timer.rs
lu-zero: Sounds correct.

src/hyperfine/benchmark.rs
lu-zero: That's a good idea.

src/hyperfine/benchmark.rs
lu-zero: Yes, probably a `format_memory` function would make that better :)

src/hyperfine/timer/unix_timer.rs
lu-zero: I wonder now if libc::wait4 wouldn't give us better information.

sharkdp:
Closing this for now. I plan to restructure the hyperfine code a little bit and will keep this use-case in mind.

lu-zero:
Great, let me know if I can help you :) 

I prepared https://crates.io/crates/wait4 a while ago to improve a bit the backend, but then I had to focus on other tasks.

sharkdp:
Thank you. This PR is not forgotten. But probably won't be merged in this form.

> I prepared https://crates.io/crates/wait4 a while ago to improve a bit the backend, but then I had to focus on other tasks.

Nice! I still don't fully understand whether it would be possible to get the memory usage of the process in question, given that we launch it through an intermediate shell (related: #336).

lu-zero:
IF we have the intermediate shell we'd have to use getrusage(RUSAGE_CHILDREN) and then subtract what wait4() is telling for the shell PID, I guess.

Worst case we'd have to know by some mean the PID of the program we care about and pass it to the wait4 or equivalent.

lu-zero:
I played a bit and I discovered that wait4(child_pid) is behaving exactly like `getrusage(RUSAGE_CHILDREN)`, that means that the shell resources would be measured and subtracted from an empty shell spawn as well.

Alex-PLACET:
@lu-zero Have you tested that on Windows ? I'm testing your branch and the result is wrong
> Memory (mean ± σ):     0.260 KiB ± 0.000 KiB
>   Range (min … max):   0.260 KiB … 0.260 KiB
For a process which takes 2 Gb

PR #327, opened 2020-10-08
scripts: modify shebang to use "/usr/bin/env python"

muxator:
Before this patch, executing directly one of these scripts, for example `./plot_hystogram.py` in a unix-like environment meant that the default system-level python would be used, regardless of an eventual activated virtualenv.

This was due to the "#!/usr/bin/python" shebang.

Changing it to "/usr/bin/env python" is a fairly standard practice, keeps intact the compatiblity with the system level python, and allows a user to run in a virtualenv if he wants.

### DIFF
--- a/scripts/advanced_statistics.py
+++ b/scripts/advanced_statistics.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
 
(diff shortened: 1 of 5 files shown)


sharkdp:

PR #330, opened 2020-10-09
add CLI option to identify a command with a custom name

### LINKED ISSUE #326: Feature request: Named commands
Hi,
Just started using hyperfine, I think it's going to find a permanent place in my toolkit.

I've been using it to test two instances of the same service running on 2 different local ports (the services in question run on remote servers that I have configured SSH tunnels to).

My invocation looks something like:
`hyperfine --export-markdown results.md 'curl localhost:1000/query' 'curl localhost:1001/query' `
The report is not great for sharing - the person(s) I share it with won't know which port refers to which instance. In a few days time I probably won't either.

I would love it if I could name the commands being ran, something like:
`hyperfine --export-markdown results.md --name 'Current platform' 'curl localhost:1000/query' --name 'New platform' 'curl localhost:1001/query' `

Cheers!

scampi:
CLI option `--name` can be used to define the name to identify a
command. If not passed, then the command itself is used. Commands
and names are paired in the same order: The first command executed gets
the first name passed as option.

Close #326

### DIFF
--- a/src/hyperfine/app.rs
+++ b/src/hyperfine/app.rs
@@ -228,7 +228,7 @@ fn build_app() -> App<'static, 'static> {
         )
         .arg(
             Arg::with_name("command-name")
-                .long("name")
+                .long("command-name")
                 .short("n")
                 .takes_value(true)
                 .multiple(true)
--- a/src/hyperfine/app.rs
+++ b/src/hyperfine/app.rs
@@ -226,6 +226,16 @@ fn build_app() -> App<'static, 'static> {
                      when trying to benchmark output speed.",
                 ),
         )
+        .arg(
+            Arg::with_name("command-name")
+                .long("name")
+                .short("n")
+                .takes_value(true)
+                .multiple(true)
+                .number_of_values(1)
+                .value_name("NAME")
+                .help("Give a meaningful name to a command"),
+        )
         .help_message("Print this help message.")
         .version_message("Show version information.")
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -130,6 +130,12 @@ fn build_hyperfine_options(matches: &ArgMatches<'_>) -> Result<HyperfineOptions,
     options.names = matches
         .values_of("command-name")
         .map(|values| values.map(String::from).collect::<Vec<String>>());
+    if let Some(ref names) = options.names {
+        let command_strings = matches.values_of("command").unwrap();
+        if names.len() > command_strings.len() {
+            return Err(OptionsError::TooManyCommandNames(command_strings.len()));
+        }
+    }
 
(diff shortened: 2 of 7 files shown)


sharkdp:
Thank you very much for your contribution! I am planning to take a closer look soon.

sharkdp:
Before I take a closer look at the code, I have a few questions:

- Does this affect the `--export-*` formats? (I think it should)
- If so, would changes to the command name also be visible when plotting data via one of the `scripts/…`?

As for the PR, could you please update the man page, if possible?

Also, could you please add an entry to the "unreleased" section in `CHANGELOG.md`? The format for entries is:
```
```

scampi:
> If so, would changes to the command name also be visible when plotting data via one of the scripts/…?

The output of the various `export`s have the custom names, and so does the chart created with a script, e.g., `plot_histogram.py`.

See new commits for the updated changelog and man page.

src/hyperfine/app.rs
```diff
@@ -226,6 +226,16 @@ fn build_app() -> App<'static, 'static> {
                      when trying to benchmark output speed.",
                 ),
         )
+        .arg(
+            Arg::with_name("command-name")
+                .long("name")
```
sharkdp, line 231: I would prefer `--command-name` for the long option. If we have a short option anyway, the long option should have a self-documenting character IMO... e.g. for shell scripts. If we change it, the man page also has to be updated.

src/main.rs
```diff
@@ -127,6 +127,10 @@ fn build_hyperfine_options(matches: &ArgMatches<'_>) -> Result<HyperfineOptions,
         (None, None) => {}
     };
 
+    options.names = matches
+        .values_of("command-name")
+        .map(|values| values.map(String::from).collect::<Vec<String>>());
```
sharkdp, line 132: I think we should show an error message if there are *more* names than commands:
```
hyperfine 'echo a' -n a -n b
```

Having too few is okay, but probably not really intended in most cases.

sharkdp (commented):
Thank you for the updates.

src/hyperfine/app.rs
scampi: sounds good, see commit https://github.com/sharkdp/hyperfine/pull/330/commits/cc6778884f1defba9fbccbef2fd8c5c0dbeaf526

src/main.rs
scampi: OK I now return an error in that case, see https://github.com/sharkdp/hyperfine/pull/330/commits/9ed5f0de40b295b5b1a670d62f439ea8dcbe98cc

sharkdp (approved):
Thank you for the updates!

PR #339, opened 2020-10-17
Export intermediate results after each benchmark command

### LINKED ISSUE #306: Writing intermediate results
Is there currently any way to force hyperfine to write intermediate results, so that I can abort a benchmark without losing all the progress I've made so far? I'm asking, because I've been benchmarking several algorithms with exponential complexity recently, and so sometimes letting the benchmark finish isn't an option.

s1ck:
fixes #306 

I took a very simple approach, by just exporting all benchmark results after each benchmark command has run. It re-creates the file, but still allows using e.g. `tail -f`.

Please let me know if this fits the requirements, happy to work on a more sophisticated implementation, e.g. opening the file and trying to append individual results, however, this becomes a bit more complicated for json I assume.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ## Changes
 
+- When `--export-*` options are used, result files are written after each individual 
+  benchmark command instead of writing after all benchmarks have finished. See #306 (@s1ck).
+
 ## Bugfixes
 
 ## Other
--- a/src/hyperfine/export/mod.rs
+++ b/src/hyperfine/export/mod.rs
@@ -69,7 +69,7 @@ impl ExportManager {
     }
 
     /// Write the given results to all Exporters contained within this manager
-    pub fn write_results(&self, results: Vec<BenchmarkResult>, unit: Option<Unit>) -> Result<()> {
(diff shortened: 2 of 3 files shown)


sharkdp:

> I took a very simple approach, by just exporting all benchmark results after each benchmark command has run. It re-creates the file

That's actually what I suggested in #313, in case you have not seen that.

> Please let me know if this fits the requirements, happy to work on a more sophisticated implementation, e.g. opening the file and trying to append individual results, however, this becomes a bit more complicated for json I assume.

I guess it's not even possible to generate always-valid JSON files this way, as they would need some sort of closing array `]` and/or record `}`.

> It re-creates the file, but still allows using e.g. `tail -f`.

The Markdown export shows relative times in one column. These relative times can change as new benchmark results come in. So using `tail -f` won't really work in that case, but that's completely fine, I guess.



Could you please add an entry to the "unreleased" section in `CHANGELOG.md`? The format for entries is:
```
```

s1ck:
Sorry, I haven't checked #313 as there was no PR linked to the issue. Reading it now and it's nice that you came to the same conclusion which is "the export should be so fast that it doesn't really matter" :)

sharkdp:
Thank you for the update!

PR #340, opened 2020-10-17
Create export file before first benchmark

s1ck:
#306 

A second request in #306 was to check if the file specified by the `--export-*` options is accessible. In this PR, we create the file when the `ExportManager` is initialized and just open it in `write` mode when writing the results.

### DIFF
--- a/src/hyperfine/export/mod.rs
+++ b/src/hyperfine/export/mod.rs
@@ -8,7 +8,7 @@ use self::csv::CsvExporter;
 use self::json::JsonExporter;
 use self::markdown::MarkdownExporter;
 
-use std::fs::File;
+use std::fs::{File, OpenOptions};
 use std::io::{Result, Write};
 
 use crate::hyperfine::types::BenchmarkResult;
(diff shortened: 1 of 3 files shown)


src/hyperfine/export/mod.rs
```diff
@@ -55,7 +55,9 @@ impl ExportManager {
     }
 
     /// Add an additional exporter to the ExportManager
-    pub fn add_exporter(&mut self, export_type: ExportType, filename: &str) {
+    pub fn add_exporter(&mut self, export_type: ExportType, filename: &str) -> Result<()> {
+        let _ = File::create(filename)?;
```
sharkdp, line 59: Another idea could be to open the file here (in write-mode, as below) and store the file handle with the exporter - instead of (or in addition to) the filename.

sharkdp:
Thank you, this looks great. I have one comment, but I'm also okay with leaving it as it is, if you prefer.

Could you please also adapt the CHANGELOG (see your other PR)?

src/hyperfine/export/mod.rs
s1ck: Good idea. If we add `file: File` to the struct, we however need to use the ExportManager as `&mut` everywhere since we need a `&mut self` in `write_results`. It works, but requires more changes. Did you think of a different way?

sharkdp: Let's leave it as it is for now - thank you!

Issue #341, opened 2020-10-17
Output improvements


sharkdp:
A few ideas from https://github.com/tdulcet/Benchmarking-Tool/issues/1: 

- [ ] Show the actual exit code if a command fails and `-i`, `--ignore-failure` is not set
- [ ] Think about adding the *median* runtime between *min* and *max*. This would be easy to implement, but I don't really like that the output width becomes even larger. Open for ideas.

PR #342, opened 2020-10-17
Add command exit code to output if it fails

KaindlJulian:
Implements the first point from #341. The exit code is included in the error message that is already being printed.

e.g.
```
$ hyperfine false
Benchmark #1: false
Error: Command terminated with non-zero exit code: 1. Use the '-i'/'--ignore-failure' option if you want to ignore this. Alternatively, use the '--show-output' option to debug what went wrong.",
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # unreleased
 
+- Add command exit code to output if it fails, see #342 (@KaindlJulian)
+
 ## Features
 
 ## Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,14 @@
 # unreleased
 
+- Add command exit code to output if it fails, see #342 (@KaindlJulian)
+
 ## Features
 
 ## Changes
 
 - When `--export-*` commands are used, result files are created before benchmark execution
-  to fail early in case of, e.g., wrong permissions. See #306 (@s1ck). 
-- When `--export-*` options are used, result files are written after each individual 
+  to fail early in case of, e.g., wrong permissions. See #306 (@s1ck).
+- When `--export-*` options are used, result files are written after each individual
   benchmark command instead of writing after all benchmarks have finished. See #306 (@s1ck).
 
 ## Bugfixes
@@ -15,8 +17,6 @@
 
 ## Packaging
 
-
-
 # v1.11.0
 
 ## Features
@@ -24,7 +24,7 @@
 - The `-L`/`--parameter-list` option can now be specified multiple times to
   evaluate all possible combinations of the listed parameters:
 
-  ``` bash
+  ```bash
   hyperfine -L number 1,2 -L letter a,b,c \
       "echo {number}{letter}" \
       "printf '%s\n' {number}{letter}"
@@ -51,7 +51,6 @@
   speed comparison, see #319
 - Show error message if `--warmup` or `--*runs` arguments can not be parsed, see #337
 - Keep output colorized when the output is not interactive and `--style=full` or `--style=color` is used.
-
 
 # v1.10.0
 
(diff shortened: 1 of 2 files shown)


src/hyperfine/benchmark.rs
```diff
@@ -61,11 +61,16 @@ pub fn time_shell_command(
...
+            exit_code = format!(": {}", code);
+        }
+
         return Err(io::Error::new(
             io::ErrorKind::Other,
-            "Command terminated with non-zero exit code. \
+            format!("Command terminated with non-zero exit code{}. \
              Use the '-i'/'--ignore-failure' option if you want to ignore this. \
-             Alternatively, use the '--show-output' option to debug what went wrong.",
+             Alternatively, use the '--show-output' option to debug what went wrong.", exit_code),
```
sharkdp, line 73: Couldn't this be implemented as:
``` rust
format!("Command terminated with non-zero exit code: {}. …", result.status.code()),
```
instead of constructing a new string first and calling `format!` twice?

sharkdp (commented):

Could you please add an entry to the "unreleased" section in `CHANGELOG.md`? The format for entries is:
```
```

src/hyperfine/benchmark.rs
KaindlJulian: Yea you're right i was just thinking if the code is `None` there should be no `": "` in the output. But i guess its safe to unwrap so I changed it to your suggestion.

CHANGELOG.md
```diff
@@ -15,16 +17,14 @@
...
-
 # v1.11.0
 
 ## Features
 
 - The `-L`/`--parameter-list` option can now be specified multiple times to
   evaluate all possible combinations of the listed parameters:
 
-  ``` bash
+  ```bash
```
sharkdp, line 27: Do you have some kind of Markdown autoformatter, or why does the whole file have changes all over the place?

src/hyperfine/benchmark.rs
```diff
@@ -63,9 +63,12 @@ pub fn time_shell_command(
     if failure_action == CmdFailureAction::RaiseError && !result.status.success() {
         return Err(io::Error::new(
             io::ErrorKind::Other,
-            "Command terminated with non-zero exit code. \
+            format!(
+                "Command terminated with non-zero exit code: {}. \
              Use the '-i'/'--ignore-failure' option if you want to ignore this. \
              Alternatively, use the '--show-output' option to debug what went wrong.",
+                result.status.code().unwrap()
```
sharkdp, line 70: I'm sorry. I didn't see that `status.code()` comes as an optional. In this case, we should avoid the `.unwrap()` here. But maybe we could do something like
```
result.status.code().map_or("?".into(), |c| format!("{}", c))
```
here?

sharkdp (changes requested):

CHANGELOG.md
KaindlJulian: Yes, I didn't even notice, my bad. Would you like me to revert the changes or are you ok with the formatting?

src/hyperfine/benchmark.rs
KaindlJulian: Probably possible yea, maybe it's better to only print:

> ...with non-zero exit code.

without the `:` when its `None`. And only print the `: [value]` when there is some.

CHANGELOG.md
sharkdp: it would be great if you could revert them.

src/hyperfine/benchmark.rs
sharkdp: That sounds great - thanks!

sharkdp: Actually, while we are at it - why not make it even better. I just read the documentation (https://doc.rust-lang.org/std/process/struct.ExitStatus.html#method.code). It says:

*"On Unix, this will return None if the process was terminated by a signal"*

So having `None` is an actual possibility here. You can easily test it by running something like
```
hyperfine "sleep 10"
```
and then calling `killall sleep` while the benchmark is running.

So maybe we should either print:
- "The command terminated with a non-zero exit code (`X`)." if the result is `Some(X)` or
- "The command has been terminated by a signal." if the result is `None`

CHANGELOG.md
KaindlJulian: :+1: reverted

sharkdp (approved):
Thank you for the update!

sharkdp:
Released in https://github.com/sharkdp/hyperfine/releases/tag/v1.12.0

PR #344, opened 2020-10-27
Add median to the end of output

nicholas-l:
I tried adding the media to between min and max however the output became rather long.

Related to #﻿341

### DIFF
--- a/src/hyperfine/benchmark.rs
+++ b/src/hyperfine/benchmark.rs
@@ -362,6 +362,7 @@ pub fn run_benchmark(
     let (mean_str, unit_mean) = format_duration_unit(t_mean, options.time_unit);
     let stddev_str = format_duration(t_stddev, Some(unit_mean));
     let min_str = format_duration(t_min, Some(unit_mean));
+    let median_str = format_duration(t_median, Some(unit_mean));
     let max_str = format_duration(t_max, Some(unit_mean));
     let num_str = format!("{} runs", t_num);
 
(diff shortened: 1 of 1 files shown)


src/hyperfine/benchmark.rs
```diff
@@ -380,10 +381,11 @@ pub fn run_benchmark(
         );
 
         println!(
-            "  Range ({} … {}):   {:>8} … {:>8}    {}",
+            "  Range ({} … {}):   {:>8} … {:>8}    (median: {})  {:>10}",
             "min".cyan(),
             "max".purple(),
             min_str.cyan(),
+            median_str.blue(),
```
sharkdp, line 388: This actually shows the median instead of the "max" entry and shows the "max" time after "median:"

nicholas-l: Oops true!

sharkdp:

To be honest, I'm not really satisfied with this solution. Thematically, the median value definitely belongs to the min/max values. It should also be displayed in a similar way, I think. Maybe we could remove the work "Range" as well as the parentheses around `(mean ± σ)` and `(min … max)` and then replace "min … max" by something that includes the median as well.

But that's just one idea. I'm open for other suggestions as well. This is more about experimenting with different options. If none of them really work, I'd rather not display the median value at all.

Sorry if that was not clear from the issue description.

nicholas-l:
Thanks for the feedback! I was trying to see what worked (see bb84f87). I'll update the PR with your suggestions.

nicholas-l:
I have updated the PR with your suggestions. I made the median yellow, as the other colours we 'taken' but can easily change it. Also I am aligning the number of runs to the end of the first line, I hope that makes sense.
```
Benchmark #1: sleep 0.3

  Time mean ± σ:       303.3 ms ±   0.6 ms    [User: 1.7 ms, System: 1.3 ms]
  min … median … max:  302.4 ms … 303.3 ms … 304.6 ms                10 runs
```

sharkdp:
I'm sorry, but the proposed layout looks much more messy than before IMO. I decided not to show the median at all for now.

PR #350, opened 2020-11-17
Add parameters to plotting scripts

b2m:
Add ~`imgname`~ `output` parameter to be able to save a plot to a file instead of showing it.
Use-case: platforms without GUI like docker containers.

Add `labels` parameter to histogram plot.
Use-case: replacing long commands with human friendly labels.

### DIFF
--- a/scripts/plot_histogram.py
+++ b/scripts/plot_histogram.py
@@ -17,7 +17,10 @@ parser.add_argument("--bins", help="Number of bins (default: auto)")
 parser.add_argument(
     "--type", help="Type of histogram (*bar*, barstacked, step, stepfilled)"
 )
-parser.add_argument("--imgname", help="Save image to the given filename.")
+parser.add_argument(
+    "-o", "--output", help="Save image to the given filename."
+)
+
 args = parser.parse_args()
 
 with open(args.file) as f:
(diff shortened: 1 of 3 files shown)


scripts/plot_histogram.py
```diff
@@ -10,16 +10,23 @@
...
 parser.add_argument("file", help="JSON file with benchmark results")
 parser.add_argument("--title", help="Plot title")
+parser.add_argument(
+    "--labels", help="Comma-separated list of entries for the plot legend"
+)
 parser.add_argument("--bins", help="Number of bins (default: auto)")
 parser.add_argument(
     "--type", help="Type of histogram (*bar*, barstacked, step, stepfilled)"
 )
+parser.add_argument("--imgname", help="Save image to the given filename.")
```
sharkdp, line 20: I would prefer if we could name this `--output` and maybe provide a `-o` shortcut. That would be in analogy to programs like `curl` or `gcc`.

sharkdp (changes requested):
This looks great - thank you.

scripts/plot_histogram.py
b2m: Completely makes sense.

sharkdp (approved):
Thank you for the update!

PR #354, opened 2020-12-16
Make --comand-name take arguments from --parameter-list and --parameter-scan

### LINKED ISSUE #351: Can we make the --command-name take the same arguments from --parameter-list as the actual command
I would like to parameterize the command name based on the parameters from the parameter list 

Currently the `--command-name` only seem to take into account the first one in the list it would be nice if we could parametrerize the `--command-name` with the `<VAR>` as we do in the actual command

Here is a concrete example I'm running
```
hyperfine "brun -c opt -- //simwork/examples:Hepatitis@run --jacHandling {jacH}  --evaluationMethod {evalMet}" --max-runs 2 --export-json hyperfine.json --parameter-list jacH DenseJacobian,SparseJacobian,ApproxJacobian --parameter-list evalMet LLVM,Haskell --command-name {jacH}-{evalMet}
Benchmark #1: {jacH}-{evalMet}
  Time (mean ± σ):     163.091 s ±  1.368 s    [User: 161.179 s, System: 1.668 s]
  Range (min … max):   162.124 s … 164.059 s    2 runs
 
Benchmark #2: brun -c opt -- //simwork/examples:Hepatitis@run --jacHandling SparseJacobian  --evaluationMethod LLVM
  Time (mean ± σ):     164.887 s ±  5.918 s    [User: 162.802 s, System: 0.919 s]
  Range (min … max):   160.703 s … 169.072 s    2 runs
 
Benchmark #3: brun -c opt -- //simwork/examples:Hepatitis@run --jacHandling ApproxJacobian  --evaluationMethod LLVM
  Time (mean ± σ):     14.131 s ±  1.073 s    [User: 12.768 s, System: 0.110 s]
  Range (min … max):   13.373 s … 14.890 s    2 runs
 
Benchmark #4: brun -c opt -- //simwork/examples:Hepatitis@run --jacHandling DenseJacobian  --evaluationMethod Haskell
  Time (mean ± σ):     481.328 s ±  0.731 s    [User: 479.976 s, System: 1.563 s]
  Range (min … max):   480.811 s … 481.845 s    2 runs
 
Benchmark #5: brun -c opt -- //simwork/examples:Hepatitis@run --jacHandling SparseJacobian  --evaluationMethod Haskell
  Time (mean ± σ):     505.488 s ±  6.620 s    [User: 503.011 s, System: 1.796 s]
  Range (min … max):   500.807 s … 510.169 s    2 runs
 
Benchmark #6: brun -c opt -- //simwork/examples:Hepatitis@run --jacHandling ApproxJacobian  --evaluationMethod Haskell
  Time (mean ± σ):     900.906 s ± 12.950 s    [User: 895.857 s, System: 3.312 s]
  Range (min … max):   891.750 s … 910.063 s    2 runs
 
Summary
  'brun -c opt -- //simwork/examples:Hepatitis@run --jacHandling ApproxJacobian  --evaluationMethod LLVM' ran
   11.54 ± 0.88 times faster than '{jacH}-{evalMet}'
   11.67 ± 0.98 times faster than 'brun -c opt -- //simwork/examples:Hepatitis@run --jacHandling SparseJacobian  --evaluationMethod LLVM'
   34.06 ± 2.59 times faster than 'brun -c opt -- //simwork/examples:Hepatitis@run --jacHandling DenseJacobian  --evaluationMethod Haskell'
   35.77 ± 2.76 times faster than 'brun -c opt -- //simwork/examples:Hepatitis@run --jacHandling SparseJacobian  --evaluationMethod Haskell'
   63.75 ± 4.93 times faster than 'brun -c opt -- //simwork/examples:Hepatitis@run --jacHandling ApproxJacobian  --evaluationMethod Haskell'
```

silathdiir:
Closes #351 

User could run the parameterized commands as below.

```
hyperfine 'echo {foo}' --parameter-list foo 1,2 --command-name name-{foo}

hyperfine 'echo {val}' --parameter-scan val 1 3 --parameter-step-size 1 --command-name name-{val}
```

### DIFF
--- a/src/hyperfine/parameter_range.rs
+++ b/src/hyperfine/parameter_range.rs
@@ -83,6 +83,7 @@ fn build_parameterized_commands<'a, T: Numeric>(
     param_min: T,
     param_max: T,
     step: T,
+    command_names: Vec<&'a str>,
     command_strings: Vec<&'a str>,
     param_name: &'a str,
 ) -> Result<Vec<Command<'a>>, ParameterScanError> {
@@ -90,9 +91,20 @@ fn build_parameterized_commands<'a, T: Numeric>(
     let param_range = RangeStep::new(param_min, param_max, step);
     let mut commands = vec![];
 
+    let mut i = 0;
+    let name_count = command_names.len();
     for value in param_range {
         for cmd in &command_strings {
+            // Sets the command name by index (remainder) if exists.
+            let name = if name_count > 0 {
+                Some(command_names[i % name_count])
+            } else {
+                None
+            };
+            i += 1;
+
             commands.push(Command::new_parametrized(
+                name,
                 cmd,
                 vec![(param_name, ParameterValue::Numeric(value.into()))],
             ));
@@ -102,10 +114,12 @@ fn build_parameterized_commands<'a, T: Numeric>(
 }
 
 pub fn get_parameterized_commands<'a>(
+    command_names: Option<Values<'a>>,
     command_strings: Values<'a>,
     mut vals: clap::Values<'a>,
     step: Option<&str>,
 ) -> Result<Vec<Command<'a>>, ParameterScanError> {
+    let command_names = command_names.map_or(vec![], |names| names.collect::<Vec<&str>>());
     let command_strings = command_strings.collect::<Vec<&str>>();
     let param_name = vals.next().unwrap();
     let param_min = vals.next().unwrap();
@@ -121,6 +135,7 @@ pub fn get_parameterized_commands<'a>(
             param_min,
             param_max,
             step,
+            command_names,
             command_strings,
             param_name,
         );
@@ -135,7 +150,14 @@ pub fn get_parameterized_commands<'a>(
     }
 
     let step = Decimal::from_str(step.unwrap())?;
-    build_parameterized_commands(param_min, param_max, step, command_strings, param_name)
+    build_parameterized_commands(
+        param_min,
+        param_max,
+        step,
+        command_names,
+        command_strings,
+        param_name,
+    )
 }
 
 #[test]
@@ -163,8 +185,9 @@ fn test_decimal_range() {
 #[test]
 fn test_get_parameterized_commands_int() {
     let commands =
-        build_parameterized_commands(1i32, 7i32, 3i32, vec!["echo {val}"], "val").unwrap();
+        build_parameterized_commands(1i32, 7i32, 3i32, vec![], vec!["echo {val}"], "val").unwrap();
     assert_eq!(commands.len(), 3);
+    assert_eq!(commands[2].get_name(), "echo 7");
     assert_eq!(commands[2].get_shell_command(), "echo 7");
 }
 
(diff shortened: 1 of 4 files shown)


src/hyperfine/parameter_range.rs
```diff
@@ -83,16 +83,28 @@ fn build_parameterized_commands<'a, T: Numeric>(
...
     let param_range = RangeStep::new(param_min, param_max, step);
     let mut commands = vec![];
 
+    let mut i = 0;
+    let name_count = command_names.len();
     for value in param_range {
         for cmd in &command_strings {
+            // Sets the command name by index (remainder) if exists.
+            let name = if name_count > 0 {
+                Some(command_names[i % name_count])
```
sharkdp, line 100: I think we should force the list of command names to have the correct size (and show an error otherwise). In that case, the modulo operation here shouldn't be necessary, right?

sharkdp:
This looks great - thank you very much!

src/hyperfine/parameter_range.rs
silathdiir: Hi @sharkdp , sorry for the late reply. I tried to remember the problem when I implemented the code.

I think I tried to implement multiple command and command name Mapping for `--parameter-list` and `--parameter-scan` as below.

```
cargo run --
  'echo bar-{bar}' 'echo foo-{foo}'
  --parameter-list bar 1,2 --parameter-list foo 1,2
  --command-name Bar-{bar} --command-name Foo-{foo}
```

It could output the right result with this `remainder` code.

```
Benchmark #1: Bar-1

Benchmark #2: Foo-1

Benchmark #3: Bar-2

Benchmark #4: Foo-1

Benchmark #5: Bar-1

Benchmark #6: Foo-2

Benchmark #7: Bar-2

Benchmark #8: Foo-2
```

But this code really brings some issues if user specifies less `--command-name` than command size with `--parameter-list` and `--parameter-scan`.
Since the `--command-name` could be ignored or less than the command size. The missing command names are displayed as the shell command directly.

```
cargo run -- 'echo 1' 'echo 2' --command-name Name-1

Benchmark #1: Name-1

Benchmark #2: echo 2
```

But with this PR if user specifies less no-parameterized `--command-name` with `--parameter-scan`, it causes the issue as below. I think the third command name should be the shell command directly (not be `Name-1`).

```
cargo run -- 'echo {bar}' --parameter-scan bar 1 3 --parameter-step-size 1 --command-name Name-1 --command-name Name-2

Benchmark #1: Name-1

Benchmark #2: Name-2

Benchmark #3: Name-1
```

What do you think? Could you help give some suggestions about this problem? Thanks.

sharkdp: Hmm. Ok. Let's try to make this a bit more formal.

In general, there can be:
- *C* commands
- *P* parameters with
- *V_m* values for each of the *P* parameters.

The total number of benchmarks B will be given by

*B = C * V_1 * V_2 * … * V_P*

Note that in your example (slightly modifiied) `'echo bar-{bar}' 'echo foo-{foo}' --parameter-list bar 1,2 --parameter-list foo 1,2,3` example, there would actually be 2 · 2 · 3 = 12 benchmarks, not just 2 · 3 = 6.

Up to now, users can specify `--command-name` up to *B* times. If the number of command names is less than *B*, the default generated name will be used. I think we should actually remove that freedom and implement the following behavior:

Either:
- `--command-name=<name>` appears exactly *B* times => use the i-th `<name>` for the i-th benchmark
Or:
- `--command-name=<name>` appears exactly once. In this case, *every* benchmark will use the `<name>`

In both cases, variable substitution will take place.

sharkdp: @sorairolake Are you still interested in working on this? (no need for a rush. I'm only asking in case we need someone else to finish this).

silathdiir: Sorry for late reply. I will continue working on this in this week.

silathdiir:
Closes this PR since clones it to PR #391

PR #369, opened 2021-02-19
Added tab-completion support for Elvish shell

ParampreetR:
Clap provides tab-completion support for 5 shells bash, fish, zsh, powershell, elvish.

Signed-off-by: parampreetR <raja.rai6789@gmail.com>


sharkdp:
Thank you for your contribution. Adding Elvish in this place will build the files, but we won't have them included in releases. The `ci/` scripts would have to be adapted for that.

PR #371, opened 2021-02-28
Export exit codes to JSON output

### LINKED ISSUE #348: Export exit codes to JSON
If running a benchmark with `--ignore-failure`, it might be interesting to know if some of the commands actually failed.

For example, consider a long running compile-time benchmark:
``` bash
hyperfine \
  --runs 3 \
  --ignore-failure \
  --parameter-list git_tag "v0.1,v0.2,v0.3,v0.4,v0.4.1,v0.5" \
  --prepare "git checkout --force {git_tag}; make clean" \
  --export-json "compile_time.json" \
  "make -j8"
```

We pass `--ignore-failure` because we don't want all benchmark results to be lost if one of the tags does not compile cleanly. However, at the end of the run, we would like to know if one of the commands actually failed.

We could add a `exit_codes: Vec<i32>` field to `BenchmarkResult` such that the exit codes would be exported to the JSON file.

JordiChauzi:
Fix #348.

I added the field `exit_codes: Vec<i32>` to the `BenchmarkResult` struct.

However, on `unix`, the returned `ExitStatus` can be `None`, when the process is terminated by a signal. In that case, I used the `signal` method result (from `ExitStatusExt` for `unix`) as exit code. This part is located in the `extract_exit_code` function in `benchmark.rs`. I tried to document the `unwrap` uses (I found other instances of `unwrap` in the library so I assumed it was ok to use it here).

### DIFF
--- a/src/hyperfine/benchmark.rs
+++ b/src/hyperfine/benchmark.rs
@@ -1,6 +1,6 @@
 use std::cmp;
 use std::io;
-use std::process::Stdio;
+use std::process::{ExitStatus, Stdio};
 
 use colored::*;
 use statistical::{mean, median, standard_deviation};
@@ -46,7 +46,7 @@ pub fn time_shell_command(
     show_output: bool,
     failure_action: CmdFailureAction,
     shell_spawning_time: Option<TimingResult>,
-) -> io::Result<(TimingResult, bool)> {
+) -> io::Result<(TimingResult, ExitStatus)> {
     let (stdout, stderr) = if show_output {
         (Stdio::inherit(), Stdio::inherit())
     } else {
@@ -88,7 +88,7 @@ pub fn time_shell_command(
             time_user,
             time_system,
         },
-        result.status.success(),
+        result.status,
     ))
 }
 
@@ -201,6 +201,26 @@ fn run_cleanup_command(
     run_intermediate_command(shell, command, show_output, error_output)
 }
 
+#[cfg(unix)]
+fn extract_exit_code(status: ExitStatus) -> i32 {
+    use std::os::unix::process::ExitStatusExt;
+
+    /* From the ExitStatus::code documentation:
+       "On Unix, this will return None if the process was terminated by a signal."
+       In that case, ExitStatusExt::signal should never return None.
+    */
+    status.code().unwrap_or_else(|| status.signal().unwrap())
+}
+
+#[cfg(not(unix))]
+fn extract_exit_code(status: ExitStatus) -> i32 {
+    /* From the ExitStatus::code documentation:
+       "On Unix, this will return None if the process was terminated by a signal."
+       On the other configurations, ExitStatus::code should never return None.
+    */
+    status.code().unwrap()
+}
+
 /// Run the benchmark for a single shell command
 pub fn run_benchmark(
     num: usize,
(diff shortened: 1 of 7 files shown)


src/hyperfine/benchmark.rs
```diff
@@ -201,6 +201,26 @@ fn run_cleanup_command(
...
+    */
+    status.code().unwrap_or_else(|| status.signal().unwrap())
+}
+
+#[cfg(not(unix))]
+fn extract_exit_code(status: ExitStatus) -> i32 {
+    /* From the ExitStatus::code documentation:
+       "On Unix, this will return None if the process was terminated by a signal."
+       On the other configurations, ExitStatus::code should never return None.
+    */
```
sharkdp, line 220: Ok, but I don't really see a guarantee for the latter part ("On the other configurations, ExitStatus::code should never return None.") in the documentation.

I think it might be best to simply return an `Option<i32>` from this function and later (in the export) convert exit codes to a string that either contains the exit code or an empty string. What do you think?

sharkdp:

This looks great. I added one inline comment.

src/hyperfine/benchmark.rs
sharkdp: @JordiChauzi Are you interested in finishing this? Otherwise, that's also fine! We can find someone else then.

JordiChauzi: Sorry about the delay. Last month was a little busy and I kind of forgot about this PR.
Anyway, I will try and implement the changes in the next days.
You are right, there is no real guarantee that a configuration may return `None` from the `ExitStatus::code` method (this is currently not the case but again, this is not a guarantee). Printing an empty string in that case makes sense. Do we also want to store `None` instead of unwrapping when the `ExitStatus::signal` method returns `None`?

JordiChauzi: The default behavior for `serde` and outputting `json` is to serialize `None` as `null`. This will create outputs that look like this:
```json
...
"exit_codes": [
        0,
        0,
        0,
        1,
        null,
        null,
        42,
        0,
        0,
        0
      ]
...
```

To me, it looks "good enough". If you still want the empty string behavior tell me so. If this is the case, I think I will have to add a custom serialize method pour the `exit_codes` field?

sharkdp: No, I like the `null` solution better. You are right. Thanks!

src/hyperfine/benchmark.rs
```diff
@@ -201,6 +201,22 @@ fn run_cleanup_command(
...
 
+#[cfg(unix)]
+fn extract_exit_code(status: ExitStatus) -> Option<i32> {
+    use std::os::unix::process::ExitStatusExt;
+
+    /* From the ExitStatus::code documentation:
+       "On Unix, this will return None if the process was terminated by a signal."
+       In that case, ExitStatusExt::signal should never return None.
+    */
+    status.code().or_else(|| status.signal())
```
sharkdp, line 212: I think this should either return the exit code (if it's there) OR the signal n + 128. This is a standard procedure, e.g. in Bash: https://tldp.org/LDP/abs/html/exitcodes.html

Returning just the signal number (e.g. `9` for `KILL`) could be confused with the actual exit code 9.

JordiChauzi: You are right! I could not easily find a constant in `libc` or in the `std` library, so I hardcoded `128` with a comment.

sharkdp (approved):

PR #388, opened 2021-05-06
benchmark.rs: Don't use # in benchmark name

### LINKED ISSUE #381: Hyperfine benchmark number (#N) interrupted wrongly by gitlab/gitlab
When including hyperfine output in commit message, gitlab/github may interrupt:

 ```
Benchmark #1: ./copy-libev nbd+unix:///?socket=/tmp/src.sock nbd+unix:///?socket=/tmp/dst.sock
  Time (mean ± σ):     940.9 ms ±  36.3 ms    [User: 80.8 ms, System: 120.0 ms]
  Range (min … max):   892.8 ms … 1005.3 ms    10 runs
```

As if this commit is related to issue #1 :-)

For example:
- https://gitlab.com/nbdkit/libnbd/-/issues/2
- https://gitlab.com/nbdkit/libnbd/-/commit/ae0bd902385719939fcb9e05854fcb46f3c48049

This is an issue with gitlab/github, but I can understand the motivation to make it easy for users to 
mention issues in commit message or wiki pages in any context.

Users can sanitize hyperfine output manually, e.g. replace "Benchmark #1:" with "Benchmark 1:",
but it would be even nicer if hyperfine would avoid the special #N syntax.

nirs:
Using `Benchmark #1` causes github and gitlab to interrupt hyperfine
results as link to issue #1. Remove the # from the benchmark number to
play better with these systems.

Fixes #381.

### DIFF
--- a/src/hyperfine/benchmark.rs
+++ b/src/hyperfine/benchmark.rs
@@ -219,7 +219,7 @@ pub fn run_benchmark(
     if options.output_style != OutputStyleOption::Disabled {
         println!(
             "{}{}: {}",
-            "Benchmark #".bold(),
+            "Benchmark ".bold(),
             (num + 1).to_string().bold(),
             &command_name
         );


sharkdp:

PR #394, opened 2021-05-28
Update README.md

bknowles:
Note that `hyperfine` no longer appears to have the argument `--export-json`, instead it is now just `--json`.


bknowles:
Whoops, sorry that was trying to parse the output from `bench ... --json`.

Nevermind.

PR #402, opened 2021-06-14
update dependency

remilauzier:
test pass on my computer


sharkdp:
Thank you. We update dependencies automatically using dependabot. See https://github.com/sharkdp/hyperfine/pull/405

PR #409, opened 2021-07-12
Require user and system time units to match report time unit

### LINKED ISSUE #408: User and system time should be in consistent time units
The "user" and "system" time should be reported in the same unit as mean, stddev, min and max. They should also follow `--time-unit`, if specified.

Example 1:
```
▶ hyperfine --runs 2 'stress --cpu 1 --timeout 1'
Benchmark 1: stress --cpu 1 --timeout 1
  Time (mean ± σ):      1.002 s ±  0.002 s    [User: 993.7 ms, System: 1.0 ms]
  Range (min … max):    1.000 s …  1.003 s    2 runs
```

Example 2:
```
▶ hyperfine --runs 2 'stress --cpu 1 --timeout 2' --time-unit millisecond
Benchmark 1: stress --cpu 1 --timeout 2
  Time (mean ± σ):     2000.4 ms ±   0.5 ms    [User: 1.983 s, System: 0.003 s]
  Range (min … max):   2000.0 ms … 2000.8 ms    2 runs
```

film42:
The stddev/mean/max/min time unit gives precedence to the `--time-unit`
command line flag and falls back to parsing. This changes the user and
system time to follow the same time unit as the rest of the report.

Closes #408

### DIFF
--- a/src/hyperfine/benchmark.rs
+++ b/src/hyperfine/benchmark.rs
@@ -385,14 +385,14 @@ pub fn run_benchmark(
     let system_mean = mean(&times_system);
 
     // Formatting and console output
-    let (mean_str, unit_mean) = format_duration_unit(t_mean, options.time_unit);
-    let stddev_str = format_duration(t_stddev, Some(unit_mean));
-    let min_str = format_duration(t_min, Some(unit_mean));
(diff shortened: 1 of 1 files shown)


sharkdp:

Issue #336, opened 2020-10-16
Add possibility to disable intermediate shell


sharkdp:
It would be great to have a way to disable the intermediate shell for very fast commands where we need a resolution that is better than 3 ms (typical standard deviation of the shell spawning time).

This new mode would obviously not be able to benchmark complex commands like `seq 1000 | factor`, but we could use ~~https://crates.io/crates/shellwords~~ https://crates.io/crates/shell-words to be able to run commands like `my_command --foo "some strong" --bar`.

xzfc:
Hyperfine could detect such simple commands automatically, without introducing an explicit command line option to disable the shell.

sharkdp:
> Hyperfine could detect such simple commands automatically, without introducing an explicit command line option to disable the shell.

I'm not sure how that would work:
```bash
hyperfine "foo"
```
Is `foo` really a normal command? Or a shell builtin? Or a shell function? I guess we could check somehow (by walking `PATH`?) if its an actual command, but we wouldn't know if there is a shell function overwriting that.

```bash
hyperfine 'foo "$HOME"'
```
What would we do in this case? Certainly looks like a normal command that is called with a single argument. Does the user want that variable to be expanded or not? If it should be expanded, we need a shell.

And then for actual, obvious shell commands:
```
hyperfine "foo > /dev/null"
hyperfine "foo $(time)"
hyperfine "foo && bar"
```
.. how would we detect that we need a shell for those? We would need to know lots of shell syntax to determine that.

xzfc:
My basic idea is to define a limited POSIX shell subset: single command, no (unescaped) shell meta-characters with some exceptions like `"" ''` quotes. Anything that we could run relatively easily preserving the shell semantics without spawning an actual shell. Any other command that doesn't belong to this subset should be spawned using an actual shell. Explicit `-S/--shell` should disable this optimization.

This way, we don't introduce new syntax. So, no new options, and no subtle differences between an actual shell and our one. Just some commands run faster and are measured more accurately than the others.

---

> `hyperfine "foo"`
> Is `foo` really a normal command? Or a shell builtin? Or a shell function? I guess we could check somehow (by walking `PATH`?) if it's an actual command, but we wouldn't know if there is a shell function overwriting that.

It isn't a shell function since there are no predefined functions or aliases in a fresh shell (check it with `sh -c 'declare -f'`).
It may be a shell builtin, but the list of shell-builtins is well-known and limited (we could hardcode them).
[...]

Issue #368, opened 2021-02-10
Windows user & system times incorrect


kuviman:
Looks like user and kernel time on Windows is calculated for the shell process only, which is only used for spawning other processes and waiting for them, resulting in zero times.

I tried running processes directly instead of using the shell by using this and this way I do get relevant results:

```rust
/// Run a Windows shell command using `cmd.exe /C`
#[cfg(windows)]
fn run_shell_command(
    stdout: Stdio,
    stderr: Stdio,
    command: &str,
    shell: &str,
) -> io::Result<std::process::Child> {
    if command.is_empty() {
        Command::new(shell)
            .arg("/C")
            .arg(command)
            .stdin(Stdio::null())
            .stdout(stdout)
            .stderr(stderr)
            .spawn()
    } else {
        let mut parts = command.split_whitespace();
        Command::new(parts.next().unwrap())
            .args(parts)
            .stdin(Stdio::null())
            .stdout(stdout)
            .stderr(stderr)
            .spawn()
    }
}
```

sharkdp:
Thank you for reporting this! That should be fixed.

Related: https://stackoverflow.com/questions/36011572/how-to-obtain-handles-for-all-children-process-of-current-process-in-windows

sharkdp:
Also related: https://github.com/sharkdp/hyperfine/issues/336

Pietruszek:
Hi! I'm a student using hyperfine for the research for my Master's thesis. I'm not quite sure my issue is related, since you said this problem should already be solved and my user and system times are actually above zero, but maybe this issue wasn't properly fixed? Or maybe I just have a very wrong understanding of what exactly do the User and System time show.

I'm currently benchmarking a C++,C# and Java implementations of one of the problems from the Computer Language Benchmarks game. The User and System times are always very small compared to the wall clock times. One of the example results that confuse me is:
```
Time (mean ± σ):     49.903 s ±  0.082 s    [User: 6.9 ms, System: 16.3 ms]
Range (min … max):   49.815 s … 50.043 s    10 runs
```
which is for a C++ program which should be spending most of its time dynamically allocating and deallocating memory. I've even done some profiling (using a sampling profiler) which confirms that this program spends at least 75% of its time in Windows kernel code. Why is the System time so low then? Shouldn't it be 3 orders of magnitude bigger? And the same for User time...

sharkdp:
[...]

PR #429, opened 2021-09-10
Implement shell bypassing

cipriancraciun:
As commented on #336, I've made an initial patch to allow one to disable the shell intermediation, and just parse the command as a list of words (perhaps with quoting).

As @sharkdp mentioned, I currently use `--shell ''` to disable the shell, but that is quite ugly UX;  however I wanted to keep the patch as small as possible.

Also, given that Rust doesn't easily allow cross-compile I can't `cargo check --target ...windows...`, so although I've tried to get the code right on Windows most likely there is a compile error.

----

As mentioned in the issue at least in terms of outputs, it seems that `hyperfine` already handles the shell overhead and subtracts it from the outputs.  Thus this patch is mostly to speed-up the overall execution by not passing through the shell.  It shouldn't change the measured values.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -184,6 +184,11 @@ fn build_app() -> App<'static> {
                 .help("Set the shell to use for executing benchmarked commands."),
         )
         .arg(
+            Arg::new("no-shell")
+                .long("no-shell")
+                .help("Execute benchmarked commands without a shell."),
+        )
+        .arg(
             Arg::new("ignore-failure")
                 .long("ignore-failure")
                 .short('i')
--- a/src/benchmark/mod.rs
+++ b/src/benchmark/mod.rs
@@ -103,6 +103,14 @@ pub fn mean_shell_spawning_time(
     style: OutputStyleOption,
     show_output: bool,
 ) -> Result<TimingResult> {
+    if let Shell::None = shell {
+        return Ok(TimingResult {
+            time_real: 0.0,
+            time_user: 0.0,
+            time_system: 0.0,
+        });
+    }
+
     const COUNT: u64 = 50;
     let progress_bar = if style != OutputStyleOption::Disabled {
         Some(get_progress_bar(
--- a/src/options.rs
+++ b/src/options.rs
@@ -24,6 +24,8 @@ pub enum Shell {
     Default(&'static str),
     /// Custom shell command specified via --shell
     Custom(Vec<String>),
+    /// No shell, each command is parsed (via `shell-words`) and directly executed
+    None,
 }
 
 impl Default for Shell {
@@ -37,6 +39,7 @@ impl fmt::Display for Shell {
         match self {
             Shell::Default(cmd) => write!(f, "{}", cmd),
             Shell::Custom(cmdline) => write!(f, "{}", shell_words::join(cmdline)),
+            Shell::None => write!(f, "[none]"),
         }
     }
 }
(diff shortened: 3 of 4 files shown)


cipriancraciun:
OK, thanks to the CI I was able to:
* fix the Windows build;  (I still can't actually test the Windows code, due to lack of Windows;)
* apply `cargo fmt`;
* squash all commits into a single one;

Pietruszek:
I confirm that `cargo check` does finish correctly on Windows! I've also run `cargo build` and can confirm that using the resulting binary with the shell disabled, allows to bypass the issue #368.

However, I agree with @sharkdp that using the empty string to disable the shell is quite ugly - especially since on Windows you actually have to use `--shell ""` when running hyperfine in `cmd.exe` but ```--shell `"`"``` if you run it in `powershell.exe` (which I normally do)... Apparently when supplying an empty string as a parameter in `powershell.exe`, you need to escape the quatation marks, which I didn't know about.

sharkdp:
@cipriancraciun are you still interested in finishing this work?

cipriancraciun:
> @cipriancraciun are you still interested in finishing this work?

Yes, but I'll have to make some time for this;  hopefully in the next week or two.

sharkdp:
Fantastic. No hurries!

cipriancraciun:
I've just pushed a new variant of the patch, one that is compatible with the current master.

The essentials are exactly like in the previous patch, except that now I've introduced `--no-shell` instead of `--shell ''`.

This time the patch is much more graceful due to the new `Shell` enum.

sharkdp:
@cipriancraciun Thank you very much! Just last week, I decided that I want to tackle this task (hyperfine without intermediate shell) myself after this had been open for some time. I started with a large-scale refactoring and cleanup that is still ongoing (see #482). Sorry if this caused a lot of conflicts for your PR.

I'm too deep into the refactoring process now and would like to finish that first. But I can then take care of merging your PR myself.

I am also planning to introduce a few breaking changes in the upcoming release of hyperfine (2.0). It could therefore make sense to think about the CLI again. Maybe we want to make the `--no-shell` variant the default?

I also have a slightly different implementation of this change in mind. Instead of having
```rust
/// Shell to use for executing benchmarked commands
#[derive(Debug)]
pub enum Shell {
    /// Default shell command
    Default(&'static str),

    /// Custom shell command specified via --shell
    Custom(Vec<String>),

    /// No shell, each command is parsed (via `shell-words`) and directly executed
    None,
}
```
I was thinking of something like a "benchmark executor" (or similar name) enum
```
pub enum Executor {
    Direct,
    ViaShell(Shell),
    Mocked(Fn() -> TimingResult),
}
```
where `Exectuor::Direct` would be the same as `Shell::None` in your implementation and `Exector::ViaShell(Shell)` would be the variant via a default or custom shell. The third variant is something that would be extremely helpful to write proper integration tests for hyperfine. The problem with that is that actual runtimes are never reliable. I can't just run `hyperfine "sleep 0.5"` in a test and expect the mean runtime to be exactly 0.5s. I can't even put a proper upper limit to the runtime because in principle, the process might be sent to sleep by the OS for a long time. `Executor::Mocked(enforced_result)` would maybe allow me to run hyperfine with a predetermined benchmark result for a specific command invocation.

sharkdp:
Closing this in favor of #487. Thank you very much for your initial work! I listed you as a co-author in #487.

PR #436, opened 2021-10-04
Parse --shell option as shell command

### LINKED ISSUE #410: --shell 'busybox sh', 'zsh --emulate posix', 'bash --posix'
It seems that `--shell` needs to be a single word -- which makes it impossible, amongst others, to test `busybox sh` or `zsh --emulate sh` or `bash --posix`.

Sure, for busybox you can symlink, and for zsh / bash, I suppose you can create a stub script (or small C wrapper) that `exec's` the real shell with prepended options. But of course **that perturbs the measurement**.

It might make sense to do simple whitespace splitting of the provided shell string. At most, we might have to handle '`\ `' (to allow for shell executables with whitespaces in their paths). Or we could try looking up the shell in `$PATH` first (like it's currently done), then if that fails, try whitespace splitting.

rhysd:
Fixes #410.

I did small sanity check for both normal case and error case:

スクリーンショット 2021-10-04 22 51 49


rhysd:
I fixed all reported clippy warnings as well.

sharkdp:
Thank you very much - very clean implementation!

PR #447, opened 2021-11-11
Add support for one run with --runs=1

avar:
Extend the --runs=N option added in d78c33b (Added options to specify
the max/exact numbers of runs., 2018-09-09) to support --runs=1
instead of dying with a usage error.

This is useful to do ad-hoc testing of your commands while they might
still have syntax errors, or just for doing one run where you don't
care about the stddev.

Before:

    $ /usr/bin/hyperfine -s basic -r 1 -L n 5,10 'sleep 0.{n}'
    Error: Number of runs below two

After (-s basic) also works:

    $ hyperfine -r 1 -L n 5,10 'sleep 0.{n}'
    Benchmark 1: sleep 0.5
      Time (abs ≡):        500.6 ms               [User: 0.6 ms, System: 0.0 ms]

    Benchmark 2: sleep 0.10
      Time (abs ≡):        100.8 ms               [User: 0.7 ms, System: 0.0 ms]

    Summary
      'sleep 0.10' ran
        4.97 ± 0.00 times faster than 'sleep 0.5'

This likewise combines correctly with -m and -M, probably not very
useful, but if you're tweaking an existing command-line:

    $ hyperfine -m 1 -M 1 -L n 5,10 'sleep 0.{n}'
    Benchmark 1: sleep 0.5
      Time (abs ≡):        500.6 ms               [User: 0.5 ms, System: 0.0 ms]

    Benchmark 2: sleep 0.10
      Time (abs ≡):        100.6 ms               [User: 0.6 ms, System: 0.0 ms]

    Summary
      'sleep 0.10' ran
        4.98 ± 0.00 times faster than 'sleep 0.5'

The "± 0.00" output in "faster than" should probably be adjusted too,
or we could keep it for consistency. I didn't implement that because
this is the first time I do anything in Rust, and I ran out of
template to copy when wanting to quickly implement this in
write_benchmark_comparison() in main.rs.

### DIFF
--- a/src/benchmark.rs
+++ b/src/benchmark.rs
@@ -391,7 +391,10 @@ pub fn run_benchmark(
     // Compute statistical quantities
     let t_num = times_real.len();
     let t_mean = mean(&times_real);
-    let t_stddev = standard_deviation(&times_real, Some(t_mean));
+    let mut t_stddev = 0.0;
+    if times_real.len() != 1 {
+        t_stddev = standard_deviation(&times_real, Some(t_mean));
+    }
     let t_median = median(&times_real);
     let t_min = min(&times_real);
     let t_max = max(&times_real);
@@ -409,7 +412,16 @@ pub fn run_benchmark(
     let user_str = format_duration(user_mean, Some(time_unit));
     let system_str = format_duration(system_mean, Some(time_unit));
 
-    if options.output_style != OutputStyleOption::Disabled {
+    if options.output_style != OutputStyleOption::Disabled && times_real.len() == 1 {
+        println!(
+            "  Time ({} ≡):        {:>8}  {:>8}     [User: {}, System: {}]",
+            "abs".green().bold(),
+            mean_str.green().bold(),
+            "        ".to_string(), // alignment
+            user_str.blue(),
+            system_str.blue()
+        );
+    } else if options.output_style != OutputStyleOption::Disabled {
         println!(
             "  Time ({} ± {}):     {:>8} ± {:>8}    [User: {}, System: {}]",
             "mean".green().bold(),
--- a/src/benchmark.rs
+++ b/src/benchmark.rs
@@ -391,10 +391,11 @@ pub fn run_benchmark(
     // Compute statistical quantities
     let t_num = times_real.len();
     let t_mean = mean(&times_real);
-    let mut t_stddev = 0.0;
-    if times_real.len() != 1 {
-        t_stddev = standard_deviation(&times_real, Some(t_mean));
-    }
+    let t_stddev = if times_real.len() > 1 {
+        Some(standard_deviation(&times_real, Some(t_mean)))
+    } else {
+        None
+    };
     let t_median = median(&times_real);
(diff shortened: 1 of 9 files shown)


sharkdp:
Thank you very much for your contribution. I initially restricted this on purpose, but I guess it's really helpful for debugging. This has been requested a few times.

Have you seen https://github.com/sharkdp/hyperfine/issues/276 and https://github.com/sharkdp/hyperfine/pull/282? I think it would be worth to go over a few discussion points before we merge this.

> The "± 0.00" output in "faster than" should probably be adjusted too,

Yes, I think so.

src/benchmark.rs
```diff
@@ -391,7 +391,10 @@ pub fn run_benchmark(
     // Compute statistical quantities
     let t_num = times_real.len();
     let t_mean = mean(&times_real);
-    let t_stddev = standard_deviation(&times_real, Some(t_mean));
+    let mut t_stddev = 0.0;
+    if *&times_real.len() != 1 {
```
sharkdp, line 395: It would be great if you could fix the clippy warning (see below). There is a similar warning below.

src/main.rs
```diff
@@ -171,17 +171,9 @@ fn build_hyperfine_options<'a>(
     }
 
     match (min_runs, max_runs) {
-        (Some(min), _) if min < 2 => {
-            // We need at least two runs to compute a variance.
-            return Err(OptionsError::RunsBelowTwo);
```
sharkdp, line 176: The `RunsBelowTwo` enum entry should be removed, I guess. Normally, the compiler should warn about this, but the OptionsError is public (should really be `pub(crate)`), so the compiler assumes this might be used by someone "outside"

avar: Fixed, thanks. No idea those were a thing, more Rust beginner mistakes...

src/benchmark.rs
avar: Fixed, thanks. No idea those were a thing, more Rust beginner mistakes...

src/main.rs
avar: Fixed, thanks!

avar:
> > The "± 0.00" output in "faster than" should probably be adjusted too,
> Yes, I think so.

Would you mind showing me how to do that? I tried that this morning and gave up on it. Surely something I can find out if I go through the full rust tutorial, but won't have time for that any time soon.

Basically I wanted to get the `max_runs` in `main.rs` down to `write_benchmark_comparison`, but it's not a `u64` like `min`, but in some option parser container type that I couldn't figure out how to unbox.

sharkdp:
Done. I decided to make the standard-deviation fields `Optional`. This way,, we are forced to deal with missing values at every single place in the code base. All of the export formats also needed to be adapted:
- For Markdown, we just omit the standard deviation
- For CSV, we still use "0" as a default to not destroy the numerical format
- For JSON, we use `null` to indicate the missing value
- For AsciiDoc, we print `± ?` for now.

mralusw:
Thanks, guys!

PR #448, opened 2021-11-11
Add --setup (-b) option, like --prepare but runs once per batch, steal -s from --style

avar:
Add --setup (-s) option, like --prepare but runs once per batch

Add a --setup option for the use-case of wanting to do one-off setup
before a set of benchmarks, not the once-per-test setup that --prepare
does. This is useful for the cases noted in the updated documentation.

Per the feedback on https://github.com/sharkdp/hyperfine/pull/448 this
new "--setup" option will steal the "-s" short option from
"--style" (initially this was called "--build" and used the
non-conflicting "-b").

I'd prefer if this and --cleanup took N number of commands, so you
could do e.g.:

    hyperfine -L rev master,next \
        -b 'git worktree add /tmp/test-{r}'  \
        -b 'make -C /tmp/test-{r} all'  \
        'make -C /tmp/test-{r} test' \
        -c 'git worktree remove /tmp/test-{r}'

I.e. a shortcut around not providing these with &&, which makes things
a bit more readable.

But the --cleanup option doesn't do that, so let's just go with what
it's doing for consistency, so for this you'll now need to do:

    hyperfine -L rev master,next \
        -b 'git worktree add /tmp/test-{r} &&
            make -C /tmp/test-{r} all'  \
        'make -C /tmp/test-{r} test' \
        -c 'git worktree remove /tmp/test-{r}'

While I'm at it add a missing test for not being able to call
"--cleanup" N times, to go with the corresponding new test to assert
the same for "--setup".

It might be a good idea for backwards compatibility to detect the use
of -s <str> as being [auto|basic|full|color|nocolor] and warn about
it:

    diff --git a/src/warnings.rs b/src/warnings.rs
    index 88322ad..3f36976 100644
    --- a/src/warnings.rs
    +++ b/src/warnings.rs
    @@ -6,6 +6,7 @@ use crate::units::Second;

     /// A list of all possible warnings
     pub enum Warnings {
    +    StyleAsSetup,
         FastExecutionTime,
         NonZeroExitCode,
         SlowInitialRun(Second),
    @@ -15,6 +16,11 @@ pub enum Warnings {
     impl fmt::Display for Warnings {
         fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
             match *self {
    +            Warnings::StyleAsSetup => write!(
    +                f,
    +                "You supplied one of -s [auto|basic|full|color|nocolor], you probably meant \
    +                 to use --style instead. The --setup option stole the -s short option from it."
    +            ),
                 Warnings::FastExecutionTime => write!(
                     f,
                     "Command took less than {:.0} ms to complete. Results might be inaccurate.",

### DIFF
--- a/doc/hyperfine.1
+++ b/doc/hyperfine.1
@@ -62,6 +62,12 @@ Perform at most \fImaxruns\fP (number) runs for each command. Default: no limit.
 Perform exactly \fIruns\fP (number) runs for each command. If this option is not specified,
 \fBhyperfine\fR automatically determines the number of runs.
 .HP
+\fB\-s\fR, \fB\-\-setup\fR \fIcmd\fP
+.IP
+Execute \fIcmd\fP once before each set of timing runs. This is useful
+for compiling software or doing other one-off setup.
+The \fB\-\-setup\fR option can only be specified once.
+.HP
 \fB\-p\fR, \fB\-\-prepare\fR \fIcmd...\fP
 .IP
 Execute \fIcmd\fP before each timing run. This is useful for clearing disk caches,
@@ -119,14 +125,15 @@ Example:
 .IP
 This performs benchmarks for 'gcc \-O2 main.cpp' and 'clang \-O2 main.cpp'.
 .HP
-\fB\-s\fR, \fB\-\-style\fR \fItype\fP
+\fB\-\-style\fR \fItype\fP
 .IP
 Set output style \fItype\fP (default: auto). Set this to 'basic' to disable output
 coloring and interactive elements. Set it to 'full' to enable all effects even
 if no interactive terminal was detected. Set this to 'nocolor' to keep the
 interactive output without any colors. Set this to 'color' to keep the colors
 without any interactive output. Set this to 'none' to disable all the output
-of the tool.
+of the tool. In hyperfine versions v0.4.0..v1.12.0 this option took the \fB\-s\fR,
+short option, which is now used by \fB\--setup\fR.
 .HP
 \fB\-S\fR, \fB\-\-shell\fR \fIshell\fP
 .IP
@@ -206,6 +213,18 @@ Export the results of a parameter scan benchmark to a markdown table:
 .nf
 \fBhyperfine\fR \fB\-\-export\-markdown\fR output.md \fB\-\-parameter-scan\fR time 1 5 'sleep {time}'
 .fi
+.RE
+.LP
+Demonstrate when each of \fB\-\-setup\fR, \fB\-\-prepare\fR, \fIcmd\fP and \fB\-\-cleanup\fR will run:
+.RS
+.nf
+\fBhyperfine\fR \fB\-L\fR n 1,2 \fB\-r\fR 2 \fB\-\-show-output\fR \\
+	\fB\-\-setup\fR 'echo setup n={n}' \\
+	\fB\-\-prepare\fR 'echo prepare={n}' \\
+	\fB\-\-cleanup\fR 'echo cleanup n={n}' \\
+	'echo command n={n}'
+.fi
+.RE
 .RE
 .SH AUTHOR
 .LP
--- a/src/app.rs
+++ b/src/app.rs
@@ -74,6 +74,20 @@ fn build_app() -> App<'static, 'static> {
                        hyperfine automatically determines the number of runs."),
         )
         .arg(
+            Arg::with_name("setup")
+                .long("setup")
+                .short("s")
+                .takes_value(true)
+                .number_of_values(1)
+                .value_name("CMD")
+                .help(
+                    "Execute CMD before each set of timing runs. This is useful for \
+                     compiling your software with the provided parameters, or to do any \
+                     other work that should happen once before a series of benchmark runs,\
+                     not every time as would happen with the --prepare option."
+                ),
+        )
+        .arg(
             Arg::with_name("prepare")
                 .long("prepare")
                 .short("p")
@@ -155,7 +169,6 @@ fn build_app() -> App<'static, 'static> {
         .arg(
             Arg::with_name("style")
                 .long("style")
-                .short("s")
                 .takes_value(true)
                 .value_name("TYPE")
                 .possible_values(&["auto", "basic", "full", "nocolor", "color", "none"])
--- a/src/benchmark.rs
+++ b/src/benchmark.rs
@@ -185,6 +185,18 @@ fn run_intermediate_command(
     })
 }
 
+/// Run the command specified by `--setup`.
+fn run_setup_command(
+    shell: &Shell,
+    command: &Option<Command<'_>>,
+    show_output: bool,
+) -> io::Result<TimingResult> {
+    let error_output = "The setup command terminated with a non-zero exit code. \
+                        Append ' || true' to the command if you are sure that this can be ignored.";
+
+    run_intermediate_command(shell, command, show_output, error_output)
+}
+
 /// Run the command specified by `--prepare`.
 fn run_preparation_command(
     shell: &Shell,
@@ -262,6 +274,12 @@ pub fn run_benchmark(
         };
         Command::new_parametrized(None, preparation_command, cmd.get_parameters().clone())
     });
+
+    // Run setup command
+    let setup_cmd = options.setup_command.as_ref().map(|setup_command| {
+        Command::new_parametrized(None, setup_command, cmd.get_parameters().clone())
+    });
+    run_setup_command(&options.shell, &setup_cmd, options.show_output)?;
 
     // Warmup phase
     if options.warmup_count > 0 {
--- a/src/main.rs
+++ b/src/main.rs
@@ -193,6 +193,8 @@ fn build_hyperfine_options<'a>(
         (None, None) => {}
     };
 
+    options.setup_command = matches.value_of("setup").map(String::from);
+
     options.preparation_command = matches
         .values_of("prepare")
         .map(|values| values.map(String::from).collect::<Vec<String>>());
--- a/src/options.rs
+++ b/src/options.rs
@@ -114,6 +114,9 @@ pub struct HyperfineOptions {
     /// Whether or not to ignore non-zero exit codes
     pub failure_action: CmdFailureAction,
 
+    /// Command to run before each batch of timing runs
+    pub setup_command: Option<String>,
+
     /// Command to run before each timing run
     pub preparation_command: Option<Vec<String>>,
 
@@ -146,6 +149,7 @@ impl Default for HyperfineOptions {
             runs: Runs::default(),
             min_time_sec: 3.0,
             failure_action: CmdFailureAction::RaiseError,
+            setup_command: None,
             preparation_command: None,
             cleanup_command: None,
             output_style: OutputStyleOption::Full,
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -70,6 +70,11 @@ impl ExecutionOrderTest {
         self
     }
 
+    fn setup(&mut self, output: &str) -> &mut Self {
+        self.arg("--setup");
+        self.command(output)
+    }
+
     fn prepare(&mut self, output: &str) -> &mut Self {
         self.arg("--prepare");
         self.command(output)
@@ -148,6 +153,22 @@ fn warmup_runs_are_executed_before_benchmarking_runs() {
 }
 
 #[test]
+fn setup_commands_are_executed_before_each_series_of_timing_runs() {
+    ExecutionOrderTest::new()
+        .arg("--runs=2")
+        .setup("setup")
+        .command("command 1")
+        .command("command 2")
+        .expect_output("setup")
+        .expect_output("command 1")
+        .expect_output("command 1")
+        .expect_output("setup")
+        .expect_output("command 2")
+        .expect_output("command 2")
+        .run();
+}
+
+#[test]
 fn prepare_commands_are_executed_before_each_timing_run() {
     ExecutionOrderTest::new()
         .arg("--runs=2")
@@ -178,6 +199,23 @@ fn cleanup_commands_are_executed_once_after_each_benchmark() {
         .expect_output("command 2")
         .expect_output("command 2")
         .expect_output("cleanup")
+        .run();
+}
+
+#[test]
+fn setup_prepare_cleanup_combined() {
+    ExecutionOrderTest::new()
+        .arg("--runs=2")
+        .setup("make")
+        .prepare("make testclean")
+        .command("make test")
+        .cleanup("make clean")
+        .expect_output("make")
+        .expect_output("make testclean")
+        .expect_output("make test")
+        .expect_output("make testclean")
+        .expect_output("make test")
+        .expect_output("make clean")
         .run();
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,13 @@
 
 ## Features
 
+- Added `--setup` (`s`) option that can be used to run `make all` or
+  similar. It runs once per set of tests, like `--cleanup` (`c`).
+
 ## Changes
+
+- The `-s` short option for `--style` has been usurped by the new
+  `--setup` option.
 
 ## Bugfixes
 


doc/hyperfine.1
```diff
@@ -207,6 +213,14 @@ Export the results of a parameter scan benchmark to a markdown table:
...
 .fi
 .RE
+.LP
+Demonstrate when each of \fB\-\-build\fR, \fB\-\-prepare\fR, \fIcmd\fP and \fB\-\-cleanup\fR will run:
+.RS
+.nf
+\fBhyperfine\fR \fB\-\fRL n 1,2 \fB\-r\fR 2 \fB\-\-show-output\fR \fB\-b\fR 'echo build n={n}' \fB\-\fRp 'echo warmup n={n}' 'echo test n={n}' \fB\-\fRc 'echo clean n={n}'
+.fi
+.RE
+.RE
```
sharkdp, line 223: Nice!

What I would really like to do at some point would be to add an execution flow graph somewhere in the documentation. Ideally, this should be done in graphviz (or similar) to make it easy to modify in the future.

src/app.rs
```diff
@@ -73,6 +73,21 @@ fn build_app() -> App<'static, 'static> {
                 .help("Perform exactly NUM runs for each command. If this option is not specified, \
                        hyperfine automatically determines the number of runs."),
         )
+        .arg(
+            Arg::with_name("build")
+                .long("build")
+                .short("b")
```
sharkdp, line 79: Concerning the name: I guess there are other uses for this besides "building" your software. How about renaming it to `--setup`? The help text can and should still refer to the build/compile example though.

sharkdp:

src/app.rs
avar: I'm agnostic on the name, and would have preferred `-setup`, but `-s` is taken for `--style` already. Less typing than:

    hyperfine --setup '...' -p '...' '<cmd>' -c '...'

So pretty much anything as long as it's got a short name is fine by me.

Just looking at https://www.thesaurus.com/browse/set%20up `--build`, `--begin` and `--erect` would all fit and have `-b` and `-e` available, maybe there's others.

I also see from the git history that `--prepare` used to be called `--setup`, but I don't know if you've got users that expected that that would be surprised, or if it's OK to just re-use it.

sharkdp: > Thinking about it a bit again I think `--begin` is a better name than

> with `build` I'd think the name immediately suggests to the user

So what are you advocating for now? :smile: `--build` or `--begin`?

I kind of like `--setup`, as it indicates that it is the opposite of `--cleanup`, which is - as far as I undertand this proposal - what the implemented new option does, right?

avar:
I'd just like to have a short option name. Would --setup be long option
only? Your pick, so if you'd like that....

src/app.rs
avar: Let's go with `--setup`, updated & stole `-s` from `--style`. As noted in the updated commit message it might be nice to print a warning on `-s basic|full|...`, but I couldn't find out how to do that.

doc/hyperfine.1
```diff
@@ -119,14 +125,15 @@ Example:
...
+\fB\-\-style\fR \fItype\fP
 .IP
 Set output style \fItype\fP (default: auto). Set this to 'basic' to disable output
 coloring and interactive elements. Set it to 'full' to enable all effects even
 if no interactive terminal was detected. Set this to 'nocolor' to keep the
 interactive output without any colors. Set this to 'color' to keep the colors
 without any interactive output. Set this to 'none' to disable all the output
-of the tool.
+of the tool. In hyperfine versions v0.4.0..v1.12.0 this option took the \fB\-s\fR,
+short option, which is now used by \fB\--setup\fR.
```
sharkdp, line 136: I'd prefer if we mention this in the `CHANGELOG.md` under (breaking) changes.

doc/hyperfine.1
```diff
@@ -207,6 +214,14 @@ Export the results of a parameter scan benchmark to a markdown table:
 \fBhyperfine\fR \fB\-\-export\-markdown\fR output.md \fB\-\-parameter-scan\fR time 1 5 'sleep {time}'
 .fi
 .RE
+.LP
+Demonstrate when each of \fB\-\-setup\fR, \fB\-\-prepare\fR, \fIcmd\fP and \fB\-\-cleanup\fR will run:
+.RS
+.nf
+\fBhyperfine\fR \fB\-\fRL n 1,2 \fB\-r\fR 2 \fB\-\-show-output\fR \fB\-s\fR 'echo setup n={n}' \fB\-\fRp 'echo warmup n={n}' 'echo test n={n}' \fB\-\fRc 'echo clean n={n}'
```
sharkdp, line 221: Can we please use the long version of the command line options here? Not everyone is familiar with all the shortcuts and the long options have a self-documenting character. The command could be split over multiple lines using `\`.

src/benchmark.rs
```diff
@@ -318,8 +341,10 @@ pub fn run_benchmark(
 
     // Determine number of benchmark runs
     let runs_in_min_time = (options.min_time_sec
-        / (res.time_real + prepare_res.time_real + shell_spawning_time.time_real))
-        as u64;
+        / (res.time_real
+            + setup_res.time_real
```
sharkdp, line 345: I don't think `setup_res.time_real` should be added here. It's ran only once before each series of benchmarks, so it shouldn't influence how many benchmark runs we can perform in the "min. time" that is reserved for benchmarks.

Setup time will obviously be there in addition, but shouldn't influence how many benchmarks we perform.

sharkdp: Note that the cleanup time is also not added here.

src/app.rs
```diff
@@ -73,6 +73,21 @@ fn build_app() -> App<'static, 'static> {
                 .help("Perform exactly NUM runs for each command. If this option is not specified, \
                        hyperfine automatically determines the number of runs."),
         )
+        .arg(
+            Arg::with_name("setup")
+                .long("setup")
+                .short("s")
+                .takes_value(true)
+                .multiple(true)
```
sharkdp, line 81: This should be removed. This way, you can get rid of the
``` rust
        if setup_command.len() != 1 {
            error("The '--setup' can only be provided once");
```
check below.

src/main.rs
```diff
@@ -197,6 +203,10 @@ fn build_hyperfine_options<'a>(
         (None, None) => {}
     };
 
+    options.setup_command = matches
+        .values_of("setup")
+        .map(|values| values.map(String::from).collect::<Vec<String>>());
```
sharkdp, line 208: Once you remove `.multiple(true)`, this can be written in analogy to che `cleanup_command` line below, i.e. with `value_of` instead of `values_of`.

src/options.rs
```diff
@@ -114,6 +114,9 @@ pub struct HyperfineOptions {
     /// Whether or not to ignore non-zero exit codes
     pub failure_action: CmdFailureAction,
 
+    /// Command to run before each batch of timing runs
+    pub setup_command: Option<Vec<String>>,
```
sharkdp, line 118: This should be just a `Option<String>`, since only one setup command can be provided.

tests/integration_tests.rs
```diff
@@ -136,6 +141,33 @@ fn warmup_runs_are_executed_before_benchmarking_runs() {
...
+#[test]
+fn setup_can_only_be_provided_once() {
+    hyperfine()
+        .arg("--runs=2")
+        .arg("--setup=setup")
+        .arg("--setup=setup-again")
+        .arg("true")
+        .assert()
+        .failure();
+}
```
sharkdp, line 153: this can probably be removed once you remove `.multiple(true)`.

tests/integration_tests.rs
```diff
@@ -170,6 +202,34 @@ fn cleanup_commands_are_executed_once_after_each_benchmark() {
...
+#[test]
+fn cleanup_can_only_be_provided_once() {
+    hyperfine()
+        .arg("--runs=2")
+        .arg("--cleanup=cleanup")
+        .arg("--cleanup=cleanup-again")
+        .arg("true")
+        .assert()
+        .failure();
+}
```
sharkdp, line 231: similar here. I don't think we need a test for this.

src/app.rs
sharkdp: > it might be nice to print a warning on `-s basic|full|...`, but I couldn't find out how to do that.

Once `options.setup_command` is just an `Option<String>`, you can do:
```
match options.setup_command {
  Some("basic") | Some("full") | … => {
    /* fail with error */
  },
  _ => {}
}
```

avar:
@sharkdp I think the latest push should address all your feedback, follows as a `git-range-diff` below:

```
1:  1bf414e ! 1:  d1ed135 Add --setup (-s) option, like --prepare but runs once per batch
    @@ Commit message
         "--style" (initially this was called "--build" and used the
         non-conflicting "-b").
     
    +    Potential future work:
    +
         I'd prefer if this and --cleanup took N number of commands, so you
         could do e.g.:
     
    @@ Commit message
                 'make -C /tmp/test-{r} test' \
                 -c 'git worktree remove /tmp/test-{r}'
     
    -    While I'm at it add a missing test for not being able to call
    -    "--cleanup" N times, to go with the corresponding new test to assert
    -    the same for "--setup".
    -
    -    It might be a good idea for backwards compatibility to detect the use
    -    of -s <str> as being [auto|basic|full|color|nocolor] and warn about
    -    it:
    -
    -        diff --git a/src/warnings.rs b/src/warnings.rs
    -        index 88322ad..3f36976 100644
    -        --- a/src/warnings.rs
    -        +++ b/src/warnings.rs
    -        @@ -6,6 +6,7 @@ use crate::units::Second;
    -
    -         /// A list of all possible warnings
    -         pub enum Warnings {
    -        +    StyleAsSetup,
    -             FastExecutionTime,
    -             NonZeroExitCode,
    -             SlowInitialRun(Second),
    -        @@ -15,6 +16,11 @@ pub enum Warnings {
    -         impl fmt::Display for Warnings {
    -             fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
    -                 match *self {
    -        +            Warnings::StyleAsSetup => write!(
    -        +                f,
    -        +                "You supplied one of -s [auto|basic|full|color|nocolor], you probably meant \
    -        +                 to use --style instead. The --setup option stole the -s short option from it."
    -        +            ),
    -                     Warnings::FastExecutionTime => write!(
    -                         f,
    -                         "Command took less than {:.0} ms to complete. Results might be inaccurate.",
    + ## CHANGELOG.md ##
    +@@
    + 
    + ## Features
    + 
    ++- Added `--setup` (`s`) option that can be used to run `make all` or
    ++  similar. It runs once per set of tests, like `--cleanup` (`c`).
    ++
    + ## Changes
    + 
    ++- The `-s` short option for `--style` has been usurped by the new
    ++  `--setup` option.
    ++
    + ## Bugfixes
    + 
    + ## Other
     
      ## doc/hyperfine.1 ##
     @@ doc/hyperfine.1: Perform at most \fImaxruns\fP (number) runs for each command. Default: no limit.
    @@ doc/hyperfine.1: Export the results of a parameter scan benchmark to a markdown
     +Demonstrate when each of \fB\-\-setup\fR, \fB\-\-prepare\fR, \fIcmd\fP and \fB\-\-cleanup\fR will run:
     +.RS
     +.nf
    -+\fBhyperfine\fR \fB\-\fRL n 1,2 \fB\-r\fR 2 \fB\-\-show-output\fR \fB\-s\fR 'echo setup n={n}' \fB\-\fRp 'echo warmup n={n}' 'echo test n={n}' \fB\-\fRc 'echo clean n={n}'
    ++\fBhyperfine\fR \fB\-L\fR n 1,2 \fB\-r\fR 2 \fB\-\-show-output\fR \\
    ++	\fB\-\-setup\fR 'echo setup n={n}' \\
    ++	\fB\-\-prepare\fR 'echo prepare={n}' \\
    ++	\fB\-\-cleanup\fR 'echo cleanup n={n}' \\
    ++	'echo command n={n}'
     +.fi
     +.RE
     +.RE
    @@ src/app.rs: fn build_app() -> App<'static, 'static> {
     +                .long("setup")
     +                .short("s")
    -+                .multiple(true)
     +                .help(
    @@ src/benchmark.rs: pub fn run_benchmark(
          });
      
     +    // Run setup command
    -+    let setup_cmd = options.setup_command.as_ref().map(|values| {
    -+        let setup_command = if values.len() == 1 {
    -+            &values[0]
    -+        } else {
    -+            &values[num]
    -+        };
    ++    // Run setup command
    ++    let setup_cmd = options.setup_command.as_ref().map(|setup_command| {
     +        Command::new_parametrized(None, setup_command, cmd.get_parameters().clone())
     +    });
    -+    let setup_res = run_setup_command(&options.shell, &setup_cmd, options.show_output)?;
    ++    run_setup_command(&options.shell, &setup_cmd, options.show_output)?;
     +
          // Warmup phase
          if options.warmup_count > 0 {
              let progress_bar = if options.output_style != OutputStyleOption::Disabled {
    -@@ src/benchmark.rs: pub fn run_benchmark(
    - 
    -     // Determine number of benchmark runs
    -     let runs_in_min_time = (options.min_time_sec
    --        / (res.time_real + prepare_res.time_real + shell_spawning_time.time_real))
    --        as u64;
    -+        / (res.time_real
    -+            + setup_res.time_real
    -+            + prepare_res.time_real
    -+            + shell_spawning_time.time_real)) as u64;
    - 
    -     let count = {
    -         let min = cmp::max(runs_in_min_time, options.runs.min);
     
      ## src/main.rs ##
    -@@ src/main.rs: fn run(
    - 
    -     let mut timing_results = vec![];
    - 
    -+    if let Some(setup_command) = &options.setup_command {
    -+        if setup_command.len() != 1 {
    -+            error("The '--setup' can only be provided once");
    -+        }
    -+    }
    -+
    -     if let Some(preparation_command) = &options.preparation_command {
    -         if preparation_command.len() > 1 && commands.len() != preparation_command.len() {
    -             error(
     @@ src/main.rs: fn build_hyperfine_options<'a>(
              (None, None) => {}
          };
      
    -+    options.setup_command = matches
    -+        .values_of("setup")
    -+        .map(|values| values.map(String::from).collect::<Vec<String>>());
    ++    options.setup_command = matches.value_of("setup").map(String::from);
     +
          options.preparation_command = matches
              .values_of("prepare")
    @@ src/options.rs: pub struct HyperfineOptions {
          pub failure_action: CmdFailureAction,
      
     +    /// Command to run before each batch of timing runs
    -+    pub setup_command: Option<Vec<String>>,
    ++    pub setup_command: Option<String>,
     +
          /// Command to run before each timing run
          pub preparation_command: Option<Vec<String>>,
    @@ tests/integration_tests.rs: fn warmup_runs_are_executed_before_benchmarking_runs
              .run();
      }
      
    -+#[test]
    -+fn setup_can_only_be_provided_once() {
    -+    hyperfine()
    -+        .arg("--runs=2")
    -+        .arg("--setup=setup")
    -+        .arg("--setup=setup-again")
    -+        .arg("true")
    -+        .assert()
    -+        .failure();
    -+}
    -+
     +#[test]
     +fn setup_commands_are_executed_before_each_series_of_timing_runs() {
     +    ExecutionOrderTest::new()
    @@ tests/integration_tests.rs: fn cleanup_commands_are_executed_once_after_each_ben
     +        .expect_output("make clean")
     +        .run();
     +}
    -+
    -+#[test]
    -+fn cleanup_can_only_be_provided_once() {
    -+    hyperfine()
    -+        .arg("--runs=2")
    -+        .arg("--cleanup=cleanup")
    -+        .arg("--cleanup=cleanup-again")
    -+        .arg("true")
    -+        .assert()
    -+        .failure();
    -+}
     +
      #[test]
      fn single_parameter_value() {
```

src/benchmark.rs
```diff
@@ -263,6 +275,13 @@ pub fn run_benchmark(
         Command::new_parametrized(None, preparation_command, cmd.get_parameters().clone())
     });
 
+    // Run setup command
+    // Run setup command
```
sharkdp, line 279: This is duplicated for some reason.

sharkdp (approved):
Thank you very much for the updates. Only one minor comment left.

sharkdp:
Thank you!

sharkdp:
Released in https://github.com/sharkdp/hyperfine/releases/tag/v1.13.0

PR #490, opened 2022-03-21
Export refactor markdown with markup

ppaulweber:
This PR provides the following changes:

* refactored Markdown-based exporter and table generation through a generic markup data processing and table generation which can be extended for upcoming markup flavors
* added appropriate tests

### DIFF
--- a/src/export/markup.rs
+++ b/src/export/markup.rs
@@ -3,44 +3,22 @@ use crate::output::format::format_duration_value;
 use crate::util::units::Unit;
 
 pub trait MarkupFormatter {
-    fn table(&self, data: &[&[&str]]) -> String {
-        let head: &[&str] = data.first().unwrap();
-        let tail = &data[1..];
+    fn table_results(&self, entries: &[BenchmarkResultWithRelativeSpeed], unit: Unit) -> String {
+        // prepare table header strings
+        let notation = format!("[{}]", unit.short_name());
+        let head: [&str; 5] = [
+            "Command",
+            &format!("Mean {}", notation),
+            &format!("Min {}", notation),
+            &format!("Max {}", notation),
+            "Relative",
+        ];
 
         // emit header
-        let table = &mut self.table_data(head);
+        let mut table = self.table_data(&head);
 
         // emit horizontal line
         table.push_str(&self.table_line(head.len()));
-
-        // emit data rows
-        for row in tail {
-            table.push_str(&self.table_data(row))
-        }
-
-        table.to_string()
-    }
-
-    fn table_data(&self, data: &[&str]) -> String;
-
-    fn table_line(&self, size: usize) -> String;
-
-    fn command(&self, size: &str) -> String;
-
-    fn results(
-        &self,
-        entries: &[BenchmarkResultWithRelativeSpeed],
-        unit: Unit,
-    ) -> Vec<Vec<String>> {
-        // prepare table header strings
-        let notation = format!("[{}]", unit.short_name());
-        let mut data: Vec<Vec<_>> = vec![vec![
-            "Command".to_string(),
-            format!("Mean {}", notation),
-            format!("Min {}", notation),
-            format!("Max {}", notation),
-            "Relative".to_string(),
-        ]];
 
         for entry in entries {
             let measurement = &entry.result;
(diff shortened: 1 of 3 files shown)


sharkdp:

I really like the idea of unifying some of the code between the different exporters. A refactoring was overdue. Also, thank you for splitting this into multiple PRs.

Before I take a more detailed look at this, I have two general remarks:
- Instead of having an enum that distinguishes between the different formats (`MarkupType`), we could consider another implementation based on traits. I imagine something like a
  ```rust
  trait TableFormatter {
    fn data(&self, data: &[&str]) -> String;
    fn line(&self, size: usize) -> String;
    // …
  }
  ```
  which could then be implemented by the various different formats:
  ```
  struct MarkdownTableFormatter;
  struct OrgModeTableFormatter;

  impl TableFormatter for MarkdownTableFormatter { /* … */ }
  impl TableFormatter for OrgModeTableFormatter { /* … */ }
  ```
  This would allow us to separate the format-specific things from the code. And we could even move them out to other files. There would be no need to repeat the `match kind { … }` statements all over the place.
- Could you have a look at the clippy suggestions and try to fix them? (see code view on this PR)

ppaulweber:
Hi @sharkdp, you're welcome. No problem, I first didn't want to make such a drastic change to your export facility and code base design, but to factor out the export behavior into a dedicated trait is for sure more maintainable. I'll update this PR with an refactor of the refactor :wink: and I'll fix the `clippy` suggestions as well.

ppaulweber:
@sharkdp I've updated the formatting by introducing a `MarkupTrait` which encapsulates all formatting logic now.
This change I've rebased into the other open PRs as well and adapted their incremental improvements accordingly to this new formatting behavior.

src/export/markup.rs
```diff
@@ -0,0 +1,77 @@
...
+
+    fn table_line(&self, size: usize) -> String;
+
+    fn command(&self, size: &str) -> String;
+
+    fn results(
+        &self,
+        entries: &[BenchmarkResultWithRelativeSpeed],
+        unit: Unit,
+    ) -> Vec<Vec<String>> {
```
sharkdp, line 34: I don't really understand why we need this function (in this form). Why do we need the intermediate `Vec<Vec<String>>`? Couldn't this function directly call `self.table_data(…)` etc.?

sharkdp (commented):
Thank you for working on this. It looks good overall. I added one question for my understanding.

src/export/markup.rs
ppaulweber: Yes, it would be possible, but the overall idea here is to introduce a separation of concerns: 
1) derive/prepare the values (in this case a `Vec<Vec<String>>`) which should be used for exporting;
2) use the values and export them in the desired format.

The advantage of this implementation would have two impacts for upcoming changes:
1) we can provide different kind of `results` creation functionalities without interfering with the actual exporting (writing of tables);
2) we can optimize the exports to derive a result and pass this prepared values just into the different exporters if there are multiple selected.

sharkdp: > Yes, it would be possible, but the overall idea here is to introduce a separation of concerns:
> 
>     1. derive/prepare the values (in this case a `Vec<Vec<String>>`) which should be used for exporting;

Ok, but then why does it need to be part of this trait? If we want actual separation of concerns, I would only move the markup-format-specific functionality into that trait. Another part of the code could then operate on a generic `dyn MarkupFormatter` and call the different `table*` commands as appropriate.

> The advantage of this implementation would have two impacts for upcoming changes:
> 
>     1. we can provide different kind of `results` creation functionalities without interfering with the actual exporting (writing of tables);

Not sure I understand...?

> we can optimize the exports to derive a result and pass this prepared values just into the different exporters if there are multiple selected.

Ok, agreed. It seems unnecessary to do it multiple times, but performance is really not a big concern here.

ppaulweber: > Ok, but then why does it need to be part of this trait? If we want actual separation of concerns, I would only move the markup-format-specific functionality into that trait. Another part of the code could then operate on a generic dyn MarkupFormatter and call the different table* commands as appropriate.

It is currently part of this trait, which we can (maybe) refactor out later in another step into another trait.
Furthermore, it IMHO makes the reading of the actual exporting functionality easier, because the separation I'm referring to is that the creation of results has a dedicated call here:
https://github.com/sharkdp/hyperfine/blob/0193cdeedaba28435402fceb0adec81e1042cdf5/src/export/markdown.rs#L40

And then some lines later the actual writing of that "formatted" data.
https://github.com/sharkdp/hyperfine/blob/0193cdeedaba28435402fceb0adec81e1042cdf5/src/export/markdown.rs#L46

> Not sure I understand...?

Hereby I'm referring to the upcoming PRs I've already prepared, to provide and support not only "relative" based summary results to export. The next PRs include another results generation called "runs" to create the appropriate table data for complete benchmark runs etc.

sharkdp:
I merged your changes in #510 and added some modifications myself. Hope that's fine with you. The major change is that I implemented the `Exporter` trait for all structs that implement the `MarkupExporter` (previously: `MarkupFormatter`) trait. This way, we can avoid code duplication in the exporter implementations.

PR #492, opened 2022-03-21
Export orgmode runs with markup

ppaulweber:
This PR introduces the following changes:

* improvement of `markup.rs` by providing a new individual timings data generation and provided exporter support for Emacs org-mode table emitting
* added new command line argument `--export-orgmode-runs <FILE>`
* ~~depends on #491~~ 
* depends on #510
* depends on #511

PR #493, opened 2022-03-21
Export markdown runs with markup

ppaulweber:
This PR introduces the following changes:

* provides individual timings of runs export as Markdown tables by reusing the `markup.rs` functionality provided in the PR https://github.com/sharkdp/hyperfine/pull/492
* adds new command line argument `--export-markdown-runs <FILE>`

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -249,6 +249,13 @@ fn build_command() -> Command<'static> {
                 .help("Export the timing summary statistics as a Emacs org-mode table to the given FILE."),
         )
         .arg(
+            Arg::new("export-orgmode-runs")
+                .long("export-orgmode-runs")
+                .takes_value(true)
+                .value_name("FILE")
+                .help("Export the timings of individual runs as a Emacs org-mode table to the given FILE."),
+        )
+        .arg(
             Arg::new("show-output")
                 .long("show-output")
                 .conflicts_with("style")
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -242,6 +242,13 @@ fn build_command() -> Command<'static> {
                 .help("Export the timing summary statistics as a Markdown table to the given FILE."),
         )
         .arg(
+            Arg::new("export-markdown-runs")
+                .long("export-markdown-runs")
+                .takes_value(true)
+                .value_name("FILE")
+                .help("Export the timings of individual runs as a Markdown table to the given FILE."),
+        )
+        .arg(
             Arg::new("export-orgmode")
                 .long("export-orgmode")
                 .takes_value(true)
--- a/src/export/markdown.rs
+++ b/src/export/markdown.rs
@@ -1,6 +1,7 @@
-use crate::export::markup::MarkupExporter;
-
 use super::markup::Alignment;
+use crate::export::{markup::MarkupExporter, BenchmarkResult, Exporter};
+use crate::util::units::Unit;
+use anyhow::Result;
 
 #[derive(Default)]
 pub struct MarkdownExporter {}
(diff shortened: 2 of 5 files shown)


sharkdp:
Thank you for this contribution. To be honest, I'm not convinced that we need this option in `hyperfine`. Could you tell me a bit more about your use case? Why do you want to look at individual timings? Do we think this will be a common use case? Or could it be implemented as a script (via the JSON export)

ppaulweber:
> Thank you for this contribution. To be honest, I'm not convinced that we need this option in `hyperfine`. Could you tell me a bit more about your use case?

@sharkdp my use case is that I'm using `hyperfine` in another project as benchmark tool which provides/generates/needs overview and runs output of the analyzed binaries which then get automatically rendered in a static page based website generator tool with `.md` and/or `.org` support.

> Why do you want to look at individual timings?

The question is here ... why not? From a user perspective if a tool provides different export formats then IMHO it should allow/provide the raw timings of the runs and the summary of the runs exporting.

> Do we think this will be a common use case? Or could it be implemented as a script (via the JSON export)

I would say yes, and of course this could be implemented in an external script but what about users who just want to use this tool and get the desired information in the export format of their choice?

Of course I/we can change this "yet another export argument" style in a more compact way similar to the new `--output` argument. How about I update this "runs" PRs by introducing only one new argument which allows exporting individual timings, summary, or both in the form of `--export-timing={summary,individual,all}` where `--export-timing=summary` would be the default setting for markup languages like MarkDown, OrgMode etc. and `--export-timing=individual` would be the default setting for raw formats like CSV or JSON. A CMD line example would be like: 

```
hyperfine \
  --export-markdown out.md \
  --export-timing=individual \
  # ...
```

sharkdp:
> > Why do you want to look at individual timings?
> 
> The question is here ... why not?

I'm being serious: I would like to know what kind of use cases exist where a human looks at individual timing results from a benchmark. One could argue that hyperfine provides a lot of statistical tools and features to avoid that exact same scenario.

This is always the hardest part of being a maintainer, but I would like to refrain from adding new features without a proper need for them. A simple "why not?" it not enough to convince me to accept the additional maintenance effort for that feature in the future.

ppaulweber:
> > > Why do you want to look at individual timings?
> > 
> > The question is here ... why not?
> 
> I'm being serious: I would like to know what kind of use cases exist where a human looks at individual timing results from a benchmark. One could argue that hyperfine provides a lot of statistical tools and features to avoid that exact same scenario.

@sharkdp I've explained you my use case above in the first paragraph, please have a look at https://github.com/sharkdp/hyperfine/pull/493#issuecomment-1133885905 again, thanks!

> This is always the hardest part of being a maintainer, but I would like to refrain from adding new features without a proper need for them. A simple "why not?" it not enough to convince me to accept the additional maintenance effort for that feature in the future.

No question about this, but as I explained already IMHO it makes more sense to support directly the export of raw data in different formats (since we already have the data structures in Rust at hand) rather than letting `hyperfine` users script additional converters.

Please can you comment on the command line argument as well? I'm referring to:
> > Of course I/we can change this "yet another export argument" style in a more compact way similar to the new --output argument. How about I update this "runs" PRs by introducing only one new argument which allows exporting individual timings, summary, or both in the form of --export-timing={summary,individual,all} where --export-timing=summary would be the default setting for markup languages like MarkDown, OrgMode etc. and --export-timing=individual would be the default setting for raw formats like CSV or JSON. A CMD line example would be like:

sharkdp:
I'm sorry, but I would like to close this for now. If more users show up in support of such an option, we can reconsider this.

PR #500, opened 2022-04-29
Allow multiple --setup arguments to be specified

matthew-eads:
Now --setup behaves identical to --prepare, except
setups run once before each batch, prepare run before each cmd

### DIFF
--- a/src/benchmark/mod.rs
+++ b/src/benchmark/mod.rs
@@ -63,23 +63,11 @@ impl<'a> Benchmark<'a> {
     }
 
     /// Run the command specified by `--setup`.
-    fn run_setup_command(
-        &self,
-        parameters: impl IntoIterator<Item = ParameterNameAndValue<'a>>,
-    ) -> Result<TimingResult> {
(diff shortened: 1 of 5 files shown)


matthew-eads:
Pausing this for now; realize that if we want multiple --setup args to work like multiple --prepares, the same should be true of --cleanup

sharkdp:
Thank you for looking into this. This sounds like a good idea!

>  setups run once before each batch, prepare run before each cmd

Yep. I recently drew a diagram to hopefully make this a bit more clear: https://github.com/sharkdp/hyperfine#detailed-benchmark-flowchart

> Pausing this for now; realize that if we want multiple --setup args to work like multiple --prepares, the same should be true of --cleanup

Yes, probably. Do you think it will lead to a lot of code duplication? Do you see any way to avoid it?

sharkdp:
@matthew-eads Any chance you will pick this up again? Otherwise, I think I'll close this for now. We can always reopen it later.

PR #509, opened 2022-05-14
Implement --output={null,pipe,<FILE>}

### LINKED ISSUE #377: consider an option to capture stdout without printing it
By default, Hyperfine will attach `Stdio::null()` to the stdout of the command it runs. Some programs, like greps, will detect this, alter how they execute and run more quickly as a result. For example, in the case of GNU grep, if it detects that stdout is `/dev/null`, then it will behave as if the `-q/--quiet` flag were present. This means GNU grep can exit as soon as a match is found. This is a perfectly legitimate optimization since its only observable behavior is its exit code, and its exit code is only influenced by whether a match exists or not (assuming no errors or interrupts occur).

What this means is that, e.g., `time grep foo file | wc -l` and `hyperfine 'grep foo file'` aren't quite doing the same thing. That's no fault of Hyperfine, however, it would be nice to have a way to easily work around it. As of today, here are some work-arounds:

* Use the `--show-output` flag, which captures stdout. But this _also_ prints stdout, which can be undesirable if there is a lot of output.
* In the case of grep specifically, use a query that is not expected to match and give Hyperfine the `--ignore-failure` flag.

A better work-around might be to have some other flag, maybe `--capture-output`, that causes Hyperfine to capture stdout but then do nothing with it.

I was motivated to write this feature request based on an errant benchmark found in the wild: https://old.reddit.com/r/commandline/comments/mibsw8/alternative_to_grep_less/gt90lmm/

tavianator:
Fixes #377.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -265,6 +265,7 @@ fn build_command() -> Command<'static> {
                      and apply certain optimizations. To avoid that, consider using \
                      '--output=pipe'.\n\n\
                      pipe: Feed the output through a pipe before discarding it.\n\n\
+                     inherit: Don't redirect the output at all (same as '--show-output').\n\n\
                      <FILE>: Write the output to the given file.",
                 ),
         )
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -253,6 +253,20 @@ fn build_command() -> Command<'static> {
                 ),
         )
         .arg(
+            Arg::new("output")
+                .long("output")
+                .conflicts_with("show-output")
+                .takes_value(true)
+                .value_name("WHERE")
+                .help(
+                    "Control where the output of the benchmark is redirected.\n\n  \
+                     null: Redirect output to /dev/null (the default).\n\n  \
+                     pipe: Feed the output through a pipe before discarding it.\n        \
+                     Useful if the benchmark special-cases /dev/null.\n\n  \
+                     <FILE>: Write the output to the given file.",
+                ),
+        )
+        .arg(
             Arg::new("command-name")
                 .long("command-name")
                 .short('n')
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -259,10 +259,12 @@ fn build_command() -> Command<'static> {
                 .takes_value(true)
                 .value_name("WHERE")
                 .help(
-                    "Control where the output of the benchmark is redirected.\n\n  \
-                     null: Redirect output to /dev/null (the default).\n\n  \
-                     pipe: Feed the output through a pipe before discarding it.\n        \
-                     Useful if the benchmark special-cases /dev/null.\n\n  \
+                    "Control where the output of the benchmark is redirected. <WHERE> can be:\n\n\
+                     null: Redirect output to /dev/null (the default). \
+                     Note that some programs like 'grep' detect when standard output is /dev/null \
+                     and apply certain optimizations. To avoid that, consider using \
+                     '--output=pipe'.\n\n\
+                     pipe: Feed the output through a pipe before discarding it.\n\n\
                      <FILE>: Write the output to the given file.",
                 ),
         )
--- a/src/error.rs
+++ b/src/error.rs
@@ -49,6 +49,6 @@ pub enum OptionsError<'a> {
     EmptyShell,
     #[error("Failed to parse '--shell <command>' expression as command line: {0}")]
     ShellParseError(shell_words::ParseError),
-    #[error("Unknown output policy '{0}'. Use ./{0} to output to a file named {0}.")]
+    #[error("Unknown output policy '{0}'. Use './{0}' to output to a file named '{0}'.")]
     UnknownOutputPolicy(String),
 }
--- a/src/error.rs
+++ b/src/error.rs
@@ -49,4 +49,6 @@ pub enum OptionsError<'a> {
     EmptyShell,
     #[error("Failed to parse '--shell <command>' expression as command line: {0}")]
     ShellParseError(shell_words::ParseError),
+    #[error("Unknown output policy '{0}'. Use ./{0} to output to a file named {0}.")]
+    UnknownOutputPolicy(String),
 }
--- a/src/options.rs
+++ b/src/options.rs
@@ -113,8 +113,8 @@ impl Default for RunBounds {
 /// How to handle the output of benchmarked commands
 #[derive(Debug, Clone, PartialEq)]
 pub enum CommandOutputPolicy {
-    /// Discard all output
-    Discard,
+    /// Redirect output to the null device
+    Null,
 
     /// Feed output through a pipe before discarding it
     Pipe,
@@ -123,19 +123,19 @@ pub enum CommandOutputPolicy {
     File(PathBuf),
 
     /// Show command output on the terminal
-    Forward,
+    Inherit,
 }
 
 impl Default for CommandOutputPolicy {
     fn default() -> Self {
-        CommandOutputPolicy::Discard
+        CommandOutputPolicy::Null
     }
 }
 
 impl CommandOutputPolicy {
     pub fn get_stdout_stderr(&self) -> io::Result<(Stdio, Stdio)> {
         let streams = match self {
-            CommandOutputPolicy::Discard => (Stdio::null(), Stdio::null()),
+            CommandOutputPolicy::Null => (Stdio::null(), Stdio::null()),
 
             // Typically only stdout is performance-relevant, so just pipe that
             CommandOutputPolicy::Pipe => (Stdio::piped(), Stdio::null()),
@@ -145,7 +145,7 @@ impl CommandOutputPolicy {
                 (file.into(), Stdio::null())
             }
 
-            CommandOutputPolicy::Forward => (Stdio::inherit(), Stdio::inherit()),
+            CommandOutputPolicy::Inherit => (Stdio::inherit(), Stdio::inherit()),
         };
 
         Ok(streams)
@@ -212,7 +212,7 @@ impl Default for Options {
             cleanup_command: None,
             output_style: OutputStyleOption::Full,
             executor_kind: ExecutorKind::default(),
-            command_output_policy: CommandOutputPolicy::Discard,
+            command_output_policy: CommandOutputPolicy::Null,
             time_unit: None,
         }
     }
@@ -269,12 +269,12 @@ impl Options {
         options.cleanup_command = matches.value_of("cleanup").map(String::from);
 
         options.command_output_policy = if matches.is_present("show-output") {
-            CommandOutputPolicy::Forward
+            CommandOutputPolicy::Inherit
         } else if let Some(output) = matches.value_of("output") {
             match output {
-                "null" => CommandOutputPolicy::Discard,
+                "null" => CommandOutputPolicy::Null,
                 "pipe" => CommandOutputPolicy::Pipe,
-                "inherit" => CommandOutputPolicy::Forward,
+                "inherit" => CommandOutputPolicy::Inherit,
                 arg => {
                     let path = PathBuf::from(arg);
                     if path.components().count() <= 1 {
@@ -284,7 +284,7 @@ impl Options {
                 }
             }
         } else {
-            CommandOutputPolicy::Discard
+            CommandOutputPolicy::Null
         };
 
         options.output_style = match matches.value_of("style") {
@@ -294,12 +294,12 @@ impl Options {
             Some("color") => OutputStyleOption::Color,
             Some("none") => OutputStyleOption::Disabled,
             _ => {
-                if options.command_output_policy != CommandOutputPolicy::Forward
-                    && atty::is(Stream::Stdout)
+                if options.command_output_policy == CommandOutputPolicy::Inherit
+                    || !atty::is(Stream::Stdout)
                 {
-                    OutputStyleOption::Full
-                } else {
                     OutputStyleOption::Basic
+                } else {
+                    OutputStyleOption::Full
                 }
             }
         };
--- a/src/options.rs
+++ b/src/options.rs
@@ -1,5 +1,7 @@
+use std::fs::File;
+use std::path::PathBuf;
 use std::process::{Command, Stdio};
-use std::{cmp, fmt};
+use std::{cmp, fmt, io};
 
 use anyhow::ensure;
 use atty::Stream;
@@ -109,10 +111,16 @@ impl Default for RunBounds {
 }
 
 /// How to handle the output of benchmarked commands
-#[derive(Debug, Clone, Copy, PartialEq)]
+#[derive(Debug, Clone, PartialEq)]
 pub enum CommandOutputPolicy {
     /// Discard all output
     Discard,
+
+    /// Feed output through a pipe before discarding it
+    Pipe,
+
+    /// Redirect output to a file
+    File(PathBuf),
 
     /// Show command output on the terminal
     Forward,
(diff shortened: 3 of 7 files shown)


tavianator:
TLDR: The new `--output=pipe` option redirects stdout through a pipe instead of to `/dev/null`.  On Linux, it's backed by `splice()`, so its overhead very competitive.

# Benchmarks

The first benchmark is `cat large`, which has its output handled by `--output=...`.
The other benchmarks are baselines:

- `cat large >/dev/null` is effectively `--output=null`
- `cat large | cat >/dev/null` is effectively `--output=pipe` on non-Linux (without `splice()`)
- `cat large | pv -q >/dev/null` is effectively `--output=pipe` on Linux (with `splice()`)

## `--output=null` (the default)

Basically the same as `>/dev/null`:

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `cat large` | 166.9 ± 2.0 | 160.6 | 168.7 | 1.00 |
| `cat large >/dev/null` | 167.5 ± 1.2 | 165.4 | 169.5 | 1.00 ± 0.01 |
| `cat large \| cat >/dev/null` | 492.8 ± 14.5 | 468.6 | 511.9 | 2.95 ± 0.09 |
| `cat large \| pv -q >/dev/null` | 347.2 ± 14.9 | 323.0 | 363.6 | 2.08 ± 0.09 |

## `--output=pipe`

`cat large` is within 2-3% of `| pv -q >/dev/null`, which makes sense as both are using `splice()`.
These methods are about 40% faster than `| cat >/dev/null`, which uses `read()/write()` through a pipe.

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `cat large` | 357.1 ± 4.4 | 348.8 | 363.5 | 2.13 ± 0.03 |
| `cat large >/dev/null` | 167.8 ± 1.4 | 165.6 | 170.9 | 1.00 |
| `cat large \| cat >/dev/null` | 474.3 ± 15.9 | 458.0 | 510.9 | 2.83 ± 0.10 |
| `cat large \| pv -q >/dev/null` | 348.2 ± 9.3 | 335.9 | 363.0 | 2.08 ± 0.06 |

## `--output=/tmp/file`

Writing to tmpfs is ~10% faster than writing to `/dev/null` through a pipe!  But slower than `splice()`, which makes sense.

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `cat large` | 436.0 ± 2.2 | 433.1 | 439.6 | 2.61 ± 0.03 |
| `cat large >/dev/null` | 166.9 ± 1.7 | 162.5 | 169.5 | 1.00 |
| `cat large \| cat >/dev/null` | 475.5 ± 20.2 | 445.8 | 504.4 | 2.85 ± 0.12 |
| `cat large \| pv -q >/dev/null` | 364.2 ± 24.2 | 319.6 | 399.8 | 2.18 ± 0.15 |

sharkdp:
This is awesome - thank you very much! Also for the benchmarks which clearly show that this is working correctly.

I have a couple of questions before I proceed with the review:
- Should we add `--output=terminal` and then treat `--show-output` as an alias for `--output=terminal`?
- Concerning the Linux-implementation of `--output=pipe` using `splice`: Could that possibly cause any confusion among users? Don't get me wrong - it's very cool that we can do this. But imagine someone is comparing benchmark results on Linux and MacOS and cannot explain the difference. I mean, we would expect OS-specific differences anyway in the implementation of a pipe. And the user would have to type `--output=pipe` explicitly. But I wonder if it's worth mentioning that in the documentation at least? Or do we want to add something like `--output=pipe` and `--output=pipe_splice`? (I'd rather not)

src/cli.rs
```diff
@@ -252,6 +252,20 @@ fn build_command() -> Command<'static> {
...
+            Arg::new("output")
+                .long("output")
+                .conflicts_with("show-output")
+                .takes_value(true)
+                .value_name("WHERE")
+                .help(
+                    "Control where the output of the benchmark is redirected.\n\n  \
+                     null: Redirect output to /dev/null (the default).\n\n  \
+                     pipe: Feed the output through a pipe before discarding it.\n        \
+                     Useful if the benchmark special-cases /dev/null.\n\n  \
```
sharkdp, line 265: I'm not sure if most users would understand the "Useful if the benchmark special-cases /dev/null" part. It would be great if we could try to make this easier somehow. Maybe by writing under `null`: "Note that some programs like 'grep' detect that the output goes to /dev/null, allowing them to enable certain optimizations. If you want to avoid that, consider using '--output=pipe'".

tavianator:
> This is awesome - thank you very much! Also for the benchmarks which clearly show that this is working correctly.

You're welcome!

> I have a couple of questions before I proceed with the review:
> 
> * Should we add `--output=terminal` and then treat `--show-output` as an alias for `--output=terminal`?

Yeah I think that makes sense for consistency.  Maybe `--output=show`/`shared`/`inherit` or something?  It's not necessarily a terminal if `hyperfine` is redirected.

I also think we should eventually add `--output=pty` (what I described [here](https://github.com/sharkdp/fd/issues/893#issuecomment-980428401)), so that benchmarks will think `isatty(stdout)` without flooding the current terminal with output.

Which reminds me, I think `--output=<FILE>` should require a path separator in `<FILE>`, so that we can add `--output=word` backwards compatibly.

> * Concerning the Linux-implementation of `--output=pipe` using `splice`: Could that possibly cause any confusion among users? Don't get me wrong - it's very cool that we can do this. But imagine someone is comparing benchmark results on Linux and MacOS and cannot explain the difference. I mean, we would expect OS-specific differences anyway in the implementation of a pipe. And the user would have to type `--output=pipe` explicitly. But I wonder if it's worth mentioning that in the documentation at least? Or do we want to add something like `--output=pipe` and `--output=pipe_splice`? (I'd rather not)

I don't see the benefit of a separate `--output=pipe_splice` -- if it's available, you should pretty much always use it over `pipe`, so why not make it the default?  There's so much that's different between macOS and Linux anyway that a direct comparison is probably not informative IMO.

src/cli.rs
tavianator: That's a good idea, I'll update the help text like that.

sharkdp:
> Yeah I think that makes sense for consistency. Maybe `--output=show`/`shared`/`inherit` or something? It's not necessarily a terminal if `hyperfine` is redirected.

Good point. I like `--output=inherit` which seems to be common terminology (e.g. being used in Rust or Python).

> I also think we should eventually add `--output=pty` (what I described [here](https://github.com/sharkdp/fd/issues/893#issuecomment-980428401)), so that benchmarks will think `isatty(stdout)` without flooding the current terminal with output.

I remember that comment. Yes, absolutely agreed.

> Which reminds me, I think `--output=<FILE>` should require a path separator in `<FILE>`, so that we can add `--output=word` backwards compatibly.

:+1: I was thinking about this edge case (*"what if a user wants to print to a file called `null`"*) but didn't post a comment because it seemed too nitpicky. But simply disallowing paths without a separator sounds like a good idea.

> I don't see the benefit of a separate `--output=pipe_splice` -- if it's available, you should pretty much always use it over `pipe`, so why not make it the default? There's so much that's different between macOS and Linux anyway that a direct comparison is probably not informative IMO.

Okay.

tavianator:
> > Yeah I think that makes sense for consistency. Maybe `--output=show`/`shared`/`inherit` or something? It's not necessarily a terminal if `hyperfine` is redirected.
> 
> Good point. I like `--output=inherit` which seems to be common terminology (e.g. being used in Rust or Python).

Done.

> > Which reminds me, I think `--output=<FILE>` should require a path separator in `<FILE>`, so that we can add `--output=word` backwards compatibly.
> 
> +1 I was thinking about this edge case (_"what if a user wants to print to a file called `null`"_) but didn't post a comment because it seemed too nitpicky. But simply disallowing paths without a separator sounds like a good idea.

Done.

src/timer/mod.rs
```diff
@@ -33,26 +43,58 @@ pub struct TimerResult {
...
+                file.as_raw_fd(),
+                None,
+                CHUNK_SIZE,
+                SpliceFFlags::empty(),
+            ) {
+                if bytes == 0 {
+                    break;
+                }
+            }
+        }
```
sharkdp, line 65: There are two potential error cases here.

* If `File::create("/dev/null")` fails (I have no idea if that can happen?), we would skip the entire `while` loop and not read from `output` at all... potentially blocking the benchmarked program?
* `splice` has been in Linux kernels since 2006. But I'm curious what would happen if we ran this on older (or somhow modified) kernels without a `splice` syscall. Would it fail during startup of hyperfine, since the corresponding wrapper in libc wouldn't be available? Or crash with `ENOSYS`? More importantly: is is something we should worry about, or do you think we can safely ignore it?

src/timer/mod.rs
```diff
@@ -33,26 +43,58 @@ pub struct TimerResult {
...
 pub fn execute_and_measure(mut command: Command) -> Result<TimerResult> {
     let wallclock_timer = WallClockTimer::start();
 
-    #[cfg(not(windows))]
-    let ((time_user, time_system), status) = {
-        let cpu_timer = self::unix_timer::CPUTimer::start();
-        let status = command.status()?;
-        (cpu_timer.stop(), status)
-    };
+    let mut child = command.spawn()?;
```
sharkdp, line 84: The new implementation (`spawn` + `wait`) looks much cleaner than what I did before (`status`) - thanks! I'm not sure if I just didn't think of this way of doing it, or if there's actually a difference (on non-Windows systems) with this implementation? Starting the timer only after spawning shouldn't lead to measurable differences, I assume?

src/options.rs
```diff
@@ -251,6 +270,19 @@ impl Options {
 
         options.command_output_policy = if matches.is_present("show-output") {
             CommandOutputPolicy::Forward
+        } else if let Some(output) = matches.value_of("output") {
+            match output {
+                "null" => CommandOutputPolicy::Discard,
+                "pipe" => CommandOutputPolicy::Pipe,
+                "inherit" => CommandOutputPolicy::Forward,
```
sharkdp, line 277: Can we please change the `CommandOutputPolicy` names to match the CLI options?

src/options.rs
```diff
@@ -262,7 +294,7 @@ impl Options {
             Some("color") => OutputStyleOption::Color,
             Some("none") => OutputStyleOption::Disabled,
             _ => {
-                if options.command_output_policy == CommandOutputPolicy::Discard
+                if options.command_output_policy != CommandOutputPolicy::Forward
```
sharkdp, line 297: Inverting this whole condition (and exchanging if and else blocks below) would make it a bit easier to read IMO:
```rust
if options.command_output_policy == CommandOutputPolicy::Forward || !atty::is(Stream::Stdout) {
  OutputStyleOption::Basic
} else {
  OutputStyleOption::Full
}
```

src/error.rs
```diff
@@ -49,4 +49,6 @@ pub enum OptionsError<'a> {
     EmptyShell,
     #[error("Failed to parse '--shell <command>' expression as command line: {0}")]
     ShellParseError(shell_words::ParseError),
+    #[error("Unknown output policy '{0}'. Use ./{0} to output to a file named {0}.")]
```
sharkdp, line 52: Minor:
```suggestion
    #[error("Unknown output policy '{0}'. Use './{0}' to output to a file named '{0}'.")]
```

src/timer/mod.rs
```diff
@@ -33,26 +43,58 @@ pub struct TimerResult {
     pub status: ExitStatus,
 }
 
+/// Discard the output of a child process.
+fn discard(output: ChildStdout) {
+    const CHUNK_SIZE: usize = 64 << 10;
```
sharkdp, line 48: Just out of interest: For `splice`, I assume this needs to be larger than a page size... but performance probably levels off at some point, so there's no gain in making this even larger?

src/timer/mod.rs
```diff
@@ -33,26 +43,58 @@ pub struct TimerResult {
...
     #[cfg(windows)]
-    let ((time_user, time_system), status) = {
-        let mut child = command.spawn()?;
-        let cpu_timer = self::windows_timer::CPUTimer::start_for_process(&child);
-        let status = child.wait()?;
+    let cpu_timer = self::windows_timer::CPUTimer::start_for_process(&child);
 
-        (cpu_timer.stop(), status)
-    };
+    if let Some(output) = output {
```
sharkdp, line 92: So.. this is only `Some(output)` if stdout was set to PIPE, correct?

src/timer/mod.rs
tavianator: > * If `File::create("/dev/null")` fails (I have no idea if that can happen?), we would skip the entire `while` loop and not read from `output` at all... potentially blocking the benchmarked program?

If anything fails here we will (silently) drop the `ChildStdout` instance which closes the pipe.  The child will get `EPIPE`/`SIGPIPE` the next time they write to stdout.  (No idea what happens on Windows though.)

Might be a good idea to at least log the error.

> * `splice` has been in Linux kernels since 2006. But I'm curious what would happen if we ran this on older (or somhow modified) kernels without a `splice` syscall. Would it fail during startup of hyperfine, since the corresponding wrapper in libc wouldn't be available? Or crash with `ENOSYS`? More importantly: is is something we should worry about, or do you think we can safely ignore it?

Rust 1.57.0 (the MSRV) has these as the minimum supported versions: https://doc.rust-lang.org/1.57.0/rustc/platform-support.html.  So I don't think we have to worry about anything so old that `splice()` is missing.  If you ran `hyperfine` on a system with too old of a `libc` it would fail to start due to the missing symbol.  If `libc` was new enough but the kernel was old, you'd get `ENOSYS`.

On the other hand, it could be blocked by `seccomp()`, or it could fail for some other reason.  Maybe we should fall back to the `read()`/`write()` loop before we give up.

src/timer/mod.rs
tavianator: There is a minor race window between `spawn()` and `start()` where the child could be running but not counted by the timer.  On a 2+ core machine it likely doesn't matter.  On a single-core machine you might get preempted for the child on `spawn()` and miss a full timeslice (a few ms) until `hyperfine` gets scheduled again.  Better safe than sorry:

```diff
diff --git a/src/timer/mod.rs b/src/timer/mod.rs
index 4a87c8c..4057318 100644
--- a/src/timer/mod.rs
+++ b/src/timer/mod.rs
@@ -81,17 +81,18 @@ fn discard(output: ChildStdout) {
 pub fn execute_and_measure(mut command: Command) -> Result<TimerResult> {
     let wallclock_timer = WallClockTimer::start();
 
-    let mut child = command.spawn()?;
-    let output = child.stdout.take();
-
     #[cfg(not(windows))]
     let cpu_timer = self::unix_timer::CPUTimer::start();
+
+    let mut child = command.spawn()?;
+
     #[cfg(windows)]
     let cpu_timer = self::windows_timer::CPUTimer::start_for_process(&child);
 
-    if let Some(output) = output {
+    if let Some(output) = child.stdout.take() {
         discard(output);
     }
+
     let status = child.wait()?;
 
     let (time_user, time_system) = cpu_timer.stop();
```

src/timer/mod.rs
tavianator: Yeah.

src/timer/mod.rs
tavianator: `splice()` seems to work for arbitrary sizes, not just pages.  The pipe buffers track byte offsets anyway for `read()` and `write()`.

The default pipe capacity in Linux is 16 pages, so 64K (except for s390x).  For fun, here's some benchmarks:

| `CHUNK_SIZE` | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| 16K | 361.8 ± 12.0 | 338.4 | 374.9 | 1.00 |
| 32K | 341.9 ± 19.1 | 299.0 | 367.3 | 1.00 |
| 64K | 341.9 ± 16.1 | 327.5 | 383.0 | 1.00 |
| 128K | 339.5 ± 14.1 | 309.6 | 356.1 | 1.00 |
| 256K | 339.9 ± 15.8 | 314.5 | 365.8 | 1.00 |

I was about to suggest changing to 128K, but then I checked `strace`:

```
splice(4, NULL, 3, NULL, 131072, 0)     = 65536
splice(4, NULL, 3, NULL, 131072, 0)     = 65536
splice(4, NULL, 3, NULL, 131072, 0)     = 65536
...
```

We can only splice 64K at a time anyway, since that's the capacity of the pipe, so bigger chunks are a waste.

sharkdp: Thank you.

sharkdp (approved):
Thank you!

PR #514, opened 2022-05-20
Remove extraneous characters from code sample

anothermh:
The first code sample in `scripts/README.md` includes `> ` on each line as an indicator that it is a shell command. GitHub includes a helpful *copy to clipboard* function demonstrated below:

The `> ` indicators are copied by this as well. When the code sample is pasted into a shell it produces an error, in bash (3.2.57), zsh (5.8) and fish (3.4.1).

It is possible to workaround this by manually copying and pasting each line without the `> ` indicators.

Removing the `> ` indicators from the code sample allows it to be copied and pasted into all three of the above-listed shells where the commands can complete successfully.

This pull request updates the code sample in `scripts/README.md` to remove the `> ` indicators.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -2,9 +2,9 @@ This folder contains scripts that can be used in combination with hyperfines `--
 
 ### Example:
 
-``` bash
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you

PR #538, opened 2022-08-08
Fix user and kernel times on Windows

clemenswasser:
# Description

By using a job object and attaching it to the spawned process before launch, we can record the total user and kernel times of it and all of its child processes. Sadly extracting the raw handle of the `main_thread_handle` is still nightly only 😢 (see: rust-lang/rust#96723) and required, in order to resume a initially suspended process. Therefore this new CPUTimer implementation is guarded behind a new `nightly` cargo feature and not enabled by default.

# Example

Before:
```
> cargo run --quiet -- 'cargo fmt'
Benchmark 1: cargo fmt
  Time (mean ± σ):     175.6 ms ±   1.1 ms    [User: 0.9 ms, System: 8.4 ms]
  Range (min … max):   173.3 ms … 177.3 ms    16 runs
```

After:
```
> cargo run --quiet -- 'cargo fmt'
Benchmark 1: cargo fmt
  Time (mean ± σ):     175.8 ms ±   1.9 ms    [User: 91.5 ms, System: 62.5 ms]
  Range (min … max):   173.5 ms … 180.7 ms    16 runs
```

Fixes: #368

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -275,6 +275,7 @@ dependencies = [
  "indicatif",
  "libc",
  "nix",
+ "once_cell",
  "predicates",
  "rand 0.8.5",
  "rust_decimal",
@@ -469,9 +470,9 @@ checksum = "830b246a0e5f20af87141b25c173cd1b609bd7779a4617d6ec582abaf90870f3"
 
 [[package]]
 name = "once_cell"
-version = "1.12.0"
+version = "1.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7709cef83f0c1f58f666e746a08b21e0085f7440fa6a29cc194d68aac97a4225"
+checksum = "074864da206b4973b84eb91683020dbefd6a8c3f0f38e054d93954e891935e4e"
 
 [[package]]
 name = "os_str_bytes"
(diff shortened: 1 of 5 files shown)


clemenswasser:
@sharkdp thanks for your comments. I'll look into them tomorrow. I already found a undocumented ntdll function NtResumeProcess, which resumes a process by HANDLE, which we could potentially use and before that I played around with a implementation using CreateToolhelp32Snapshot for finding the main thread, but using NtResumeProcess should definitely be better in terms of performance and code size 😁.

clemenswasser:
With the latest commit it works on stable using the undocumented but well known `ntdll.dll!NtResumeThread` which exists since Windows XP. When using nightly the feature `windows_process_extensions_main_thread_handle` can be set to use the unstable feature with the same name. The dependency on once_init is not a problem, since clap was already using it anyway.
Also I don't know why the `ActiveProcesses == 0` assertion triggered on https://github.com/sharkdp/hyperfine/runs/7864408316...
For me locally this never triggered for simple test cases like `cargo run -- 'cargo fmt'`

clemenswasser:
@sharkdp ping. Can I do something to help merge this?

sharkdp (approved):
No - this is great! Thank you very much.

PR #543, opened 2022-08-25
Fix default-output colour handling and Shell functions and aliases README sexion 

nabijaczleweli:


sharkdp:
Thank you for your contribution. This mixes different changes in a single PR. Can we please focus on one aspect per PR?

PR #554, opened 2022-09-02
#543: 2

nabijaczleweli:


README.md
```diff
@@ -28,19 +28,19 @@ A command-line benchmarking tool.
 
 To run a benchmark, you can simply call `hyperfine <command>...`. The argument(s) can be any
 shell command. For example:
-``` bash
+```sh
```
sharkdp, line 31: what is the advantage?

README.md
```diff
@@ -110,30 +110,29 @@ hyperfine -N 'grep TODO /home/user'
 
 ### Shell functions and aliases
 
-If you are using bash, you can export shell functions to directly benchmark them with hyperfine:
+Bash supports saving functions through the environment:
```
sharkdp, line 113: Why do you think this is better?

README.md
nabijaczleweli: No bash extensions used – indicate that clearly and in contrast to the parts where they are.

README.md
nabijaczleweli: Impersonal, explicitly lists what the mechanism is (you can't "export functions directly", because you export to the environment block which contains strings, not functions), and explicitly notes that this requires bash on both sides.

README.md
```diff
@@ -110,30 +110,29 @@ hyperfine -N 'grep TODO /home/user'
 
 ### Shell functions and aliases
 
-If you are using bash, you can export shell functions to directly benchmark them with hyperfine:
+Bash supports passing functions through the environment:
```
sharkdp, line 113: To be honest, I liked the previous sentence better.

README.md
```diff
@@ -110,30 +110,29 @@ hyperfine -N 'grep TODO /home/user'
...
 
-```
-$ my_function() { sleep 1; }
-$ export -f my_function
-$ hyperfine my_function
+```bash
+my_function() { sleep 1; }
+export -f my_function
+hyperfine --shell=bash my_function
```
sharkdp, line 118: :+1:

PR #555, opened 2022-09-02
#543: 1

nabijaczleweli:


src/options.rs
```diff
@@ -298,6 +298,14 @@ impl Options {
                     || !atty::is(Stream::Stdout)
                 {
                     OutputStyleOption::Basic
+                } else if !(env::var_os("TERM")
+                    .map(|t| t != "unknown" && t != "dumb")
+                    .unwrap_or(false)
+                    && env::var_os("NO_COLOR")
+                        .map(|t| t.is_empty())
+                        .unwrap_or(true))
```
sharkdp, line 306: I think the following would be a bit easier to read compared to the double-negation above, what do you think?

```suggestion
                } else if (env::var_os("TERM")
                    .map(|t| t == "unknown" || t == "dumb")
                    .unwrap_or(true)
                    || env::var_os("NO_COLOR")
                        .map(|t| !t.is_empty())
                        .unwrap_or(false))
```

sharkdp (changes requested):

PR #556, opened 2022-09-02
#543: 3

nabijaczleweli:


sharkdp:
To be honest, I fail to see why the new version is better?

nabijaczleweli:
To quote the commit message:
1. whether they're filesystem or not doesn't matter; indeed, this is easiy misleading given that most filesystems cache both file data and pages at different levels, and if you don't touch the filesystem then the caches are usually just of blockdev pages
2. the architecture doesn't matter, not only programs cause interference and load differences

Unless you only limit hyperfine to be run on i686, and only on regular files, and only on systems that don't have the capability to have kernel threads, then these hold. Otherwise, they're misleading.

PR #579, opened 2022-10-19
Implement slowness checking and --reference flag to use a command as a reference when printing or exporting summary

### LINKED ISSUE #577: Allow choice of reference benchmark
When running `hyperfine`, we get a nice output.

For example, when running `hyperfine` like this:

    hyperfine -L shell sh,ash,ksh,bash,zsh,csh,fish,tcsh,nu,pwsh '{shell} -c "echo \"Hello, world!\""'

The Summary of the output looks like this:

```
Summary
  'sh -c "echo \"Hello, world!\""' ran
    1.23 ± 0.73 times faster than 'ash -c "echo \"Hello, world!\""'
    1.90 ± 0.45 times faster than 'ksh -c "echo \"Hello, world!\""'
    2.03 ± 1.18 times faster than 'bash -c "echo \"Hello, world!\""'
    2.41 ± 0.55 times faster than 'zsh -c "echo \"Hello, world!\""'
    6.48 ± 1.29 times faster than 'csh -c "echo \"Hello, world!\""'
    7.26 ± 1.62 times faster than 'fish -c "echo \"Hello, world!\""'
    7.93 ± 1.68 times faster than 'tcsh -c "echo \"Hello, world!\""'
   26.55 ± 5.52 times faster than 'nu -c "echo \"Hello, world!\""'
  674.56 ± 131.91 times faster than 'pwsh -c "echo \"Hello, world!\""'
```

It would be really great if it were possible to pick one of the benchmark runs as "reference". For example, if I'm interested in how `nu` performs in comparison, I would like it to be on top, and the other benchmarks should be listed as `times faster` or `times slower`, depending on their performance.

In this example, when I run it like this:

    hyperfine -L shell sh,ash,ksh,bash,zsh,csh,fish,tcsh,nu,pwsh --reference nu '{shell} -c "echo \"Hello, world!\""'

Then the Summary should look like this:
Summary
```
  'nu -c "echo \"Hello, world!\""' ran
    26.55 ± 5.52 times slower than 'sh -c "echo \"Hello, world!\""'
    21.58 ± 7.56 times slower than 'ash -c "echo \"Hello, world!\""'
    13.97 ± 12.27 times slower than 'ksh -c "echo \"Hello, world!\""'
    13.08 ± 4.68 times slower than 'bash -c "echo \"Hello, world!\""'
    11.02 ± 10.04 times slower than 'zsh -c "echo \"Hello, world!\""'
    4.10 ± 4.27 times slower than 'csh -c "echo \"Hello, world!\""'
    3.66 ± 3.40 times slower than 'fish -c "echo \"Hello, world!\""'
    3.35 ± 3.29 times slower than 'tcsh -c "echo \"Hello, world!\""'
   25.41 ± 23.90 times faster than 'pwsh -c "echo \"Hello, world!\""'
```

Animeshz:
As title enlists.

Closes #577.

### DIFF
--- a/src/benchmark/relative_speed.rs
+++ b/src/benchmark/relative_speed.rs
@@ -8,20 +8,26 @@ pub struct BenchmarkResultWithRelativeSpeed<'a> {
     pub result: &'a BenchmarkResult,
     pub relative_speed: Scalar,
     pub relative_speed_stddev: Option<Scalar>,
-    pub is_fastest: bool,
+    pub is_reference: bool,
+    pub is_faster: bool,
 }
 
 pub fn compare_mean_time(l: &BenchmarkResult, r: &BenchmarkResult) -> Ordering {
     l.mean.partial_cmp(&r.mean).unwrap_or(Ordering::Equal)
 }
 
-pub fn compute(results: &[BenchmarkResult]) -> Option<Vec<BenchmarkResultWithRelativeSpeed>> {
-    let fastest: &BenchmarkResult = results
+pub fn fastest(results: &[BenchmarkResult]) -> &BenchmarkResult {
+    results
         .iter()
         .min_by(|&l, &r| compare_mean_time(l, r))
-        .expect("at least one benchmark result");
+        .expect("at least one benchmark result")
+}
 
-    if fastest.mean == 0.0 {
+pub fn compute<'a>(
+    reference: &'a BenchmarkResult,
+    results: &'a [BenchmarkResult],
+) -> Option<Vec<BenchmarkResultWithRelativeSpeed<'a>>> {
+    if reference.mean == 0.0 {
         return None;
     }
 
@@ -29,25 +35,34 @@ pub fn compute(results: &[BenchmarkResult]) -> Option<Vec<BenchmarkResultWithRel
         results
             .iter()
             .map(|result| {
-                let ratio = result.mean / fastest.mean;
+                let is_reference = result == reference;
+                let is_faster = result.mean >= reference.mean;
 
-                // https://en.wikipedia.org/wiki/Propagation_of_uncertainty#Example_formulas
-                // Covariance asssumed to be 0, i.e. variables are assumed to be independent
-                let ratio_stddev = match (result.stddev, fastest.stddev) {
-                    (Some(result_stddev), Some(fastest_stddev)) => Some(
-                        ratio
-                            * ((result_stddev / result.mean).powi(2)
-                                + (fastest_stddev / fastest.mean).powi(2))
-                            .sqrt(),
-                    ),
-                    _ => None,
+                let ratio = if is_faster {
+                    result.mean / reference.mean
+                } else {
+                    reference.mean / result.mean
                 };
+
+                // https://en.wikipedia.org/wiki/Propagation_of_uncertainty#Example_formulae
+                // Covariance asssumed to be 0, i.e. variables are assumed to be independent
+                let ratio_stddev =
+                    (result.stddev)
+                        .zip(reference.stddev)
+                        .map(|(result_stddev, fastest_stddev)| {
+                            let (a, b) = (
+                                (result_stddev / result.mean),
+                                (fastest_stddev / reference.mean),
+                            );
+                            ratio * (a.powi(2) + b.powi(2)).sqrt()
+                        });
 
                 BenchmarkResultWithRelativeSpeed {
                     result,
                     relative_speed: ratio,
                     relative_speed_stddev: ratio_stddev,
-                    is_fastest: result == fastest,
+                    is_reference,
+                    is_faster,
                 }
             })
             .collect(),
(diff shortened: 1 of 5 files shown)


Animeshz:
Example run:

```bash
$ hyperfine -N --reference 'bash -c "echo \"Hello, World!\""' -L shell dash,fish '{shell} -c "echo \"Hello, world!\""'
    Finished dev [unoptimized + debuginfo] target(s) in 0.08s
     Running `target/debug/hyperfine -N --reference 'bash -c "echo \"Hello, World'\!'\""' -L shell dash,fish '{shell} -c "echo \"Hello, world'\!'\""'`
Benchmark 1: bash -c "echo \"Hello, World!\""
  Time (mean ± σ):       4.0 ms ±   0.1 ms    [User: 2.8 ms, System: 1.1 ms]
  Range (min … max):     3.7 ms …   4.9 ms    677 runs
 
Benchmark 2: dash -c "echo \"Hello, world!\""
  Time (mean ± σ):       1.1 ms ±   0.1 ms    [User: 0.3 ms, System: 0.7 ms]
  Range (min … max):     0.9 ms …   1.6 ms    2489 runs
 
Benchmark 3: fish -c "echo \"Hello, world!\""
  Time (mean ± σ):      56.2 ms ±   0.8 ms    [User: 41.6 ms, System: 20.1 ms]
  Range (min … max):    54.4 ms …  58.2 ms    54 runs
 
Summary
  'bash -c "echo \"Hello, World!\""' ran
    3.72 ± 0.25 times slower than 'dash -c "echo \"Hello, world!\""'
   14.06 ± 0.42 times faster than 'fish -c "echo \"Hello, world!\""'

```

Animeshz:
@sharkdp

(just for ping)

sharkdp:

> As title enlists.

What does "slowness checking" mean?

src/benchmark/relative_speed.rs
```diff
@@ -8,46 +8,61 @@ pub struct BenchmarkResultWithRelativeSpeed<'a> {
     pub result: &'a BenchmarkResult,
     pub relative_speed: Scalar,
     pub relative_speed_stddev: Option<Scalar>,
-    pub is_fastest: bool,
+    pub is_reference: bool,
+    pub is_faster: bool,
```
sharkdp, line 12: What's the new meaning here? Maybe rename to `is_faster_than_xyz`, with `xyz` replaced by what ever it is faster than? `is_faster_than_reference`?

sharkdp, line 12: Maybe rename the latter to
```suggestion
    pub is_reference: bool,
    pub is_faster_than_reference: bool,
```

Thinking about this some more, it's actually not clear what the value of the latter bool should be for the case where `is_reference` is `true`. So maybe introduce an enum instead `pub enum RelativeOrdering { Reference, FasterThanReference, SlowerThanReference }` and use this instead of the two bool ("make illegal states unrepresentable"):
```suggestion
    pub relative_ordering: RelativeOrdering,
```

What do you think?

src/benchmark/scheduler.rs
```diff
@@ -40,7 +40,12 @@ impl<'a> Scheduler<'a> {
 
         executor.calibrate()?;
 
-        for (number, cmd) in self.commands.iter().enumerate() {
+        let reference = self
+            .options
+            .reference_command
+            .as_ref()
+            .map(|cmd| Command::new(None, &cmd));
```
sharkdp, line 47:
```suggestion
            .map(|cmd| Command::new(None, cmd));
```

src/cli.rs
```diff
@@ -68,6 +68,15 @@ fn build_command() -> Command<'static> {
...
                        hyperfine automatically determines the number of runs."),
         )
+        .arg(
+            Arg::new("reference")
+                .long("reference")
+                .short('R')
+                .takes_value(true)
+                .number_of_values(1)
+                .value_name("CMD")
+                .help("The reference command to measure the results against."),
```
sharkdp, line 78:
```suggestion
                .help("The reference command for the relative comparison of results. Default: the fastest command."),
```

src/options.rs
```diff
@@ -178,6 +178,9 @@ pub struct Options {
     /// Whether or not to ignore non-zero exit codes
     pub command_failure_action: CmdFailureAction,
 
+    /// Command to run before each *batch* of timing runs, i.e. before each individual benchmark
```
sharkdp, line 181: Copy & paste error?

src/cli.rs
```diff
@@ -68,6 +68,15 @@ fn build_command() -> Command<'static> {
                 .help("Perform exactly NUM runs for each command. If this option is not specified, \
                        hyperfine automatically determines the number of runs."),
         )
+        .arg(
+            Arg::new("reference")
+                .long("reference")
+                .short('R')
```
sharkdp, line 74: Let's not introduce a short command line option for this right away. We can always add it later if this is really a commonly used option.

sharkdp:
Closing this in favor of #744

PR #581, opened 2022-10-21
Add elapsed time to progress bar

### LINKED ISSUE #416: Request: Add ticking elapsed time on "initial time measurement"
It would be helpful to see the elapsed time on the _initial_ time measurement.

Right now, I have a very long running benchmark subject sitting at "Initial time measurement" and "ETA 00:00:00"; I'd love to see the elapsed time ticking at least in this case, e.g. as " AT 00:10:23", indicating that hyperfine is 10 minutes 23 seconds into the initial run.

devonhollowood:
I think it's pretty reasonable to have this display by default. If you want to hide it behind an option, though, or you want me to change the format / wording, please let me know!

Fixes #416.

### DIFF
--- a/src/benchmark/mod.rs
+++ b/src/benchmark/mod.rs
@@ -10,7 +10,9 @@ use crate::command::Command;
 use crate::options::{CmdFailureAction, ExecutorKind, Options, OutputStyleOption};
 use crate::outlier_detection::{modified_zscores, OUTLIER_THRESHOLD};
 use crate::output::format::{format_duration, format_duration_unit};
-use crate::output::progress_bar::get_progress_bar;
+use crate::output::progress_bar::{
+    get_progress_bar, replace_message_template, reset_progress_template,
+};
 use crate::output::warnings::Warnings;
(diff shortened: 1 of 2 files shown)


sharkdp:

Hm. I think the idea of the ticket was to only show this for the very first run. I'm not sure about the added value if we show this always. I think less information might be the better default here. Maybe we could simply replace the "Initial time measurement" message with: "Initial time measurement: {elapsed}"? would that work?

devonhollowood:
> Maybe we could simply replace the "Initial time measurement" message with: "Initial time measurement: {elapsed}"? would that work?

This doesn't work, because template arguments in the `ProgressBar`'s `message` don't get substituted. So we need a slightly deeper change. I'll see what I can come up with.

devonhollowood:
Okay, I think this should work. Here's a gif of what it looks like.
hyperfine-time-elapsed

cblair:
Still something that would be useful, I was suprised hyperfine didn't have this already.

PR #582, opened 2022-10-22
Hyperfine no longer auto assumes /C for windows if cmd.exe is not the selected shell.

### LINKED ISSUE #568: hyperfine not working with nushell
Using hyperfine on Windows with nushell doesn't work properly.

Trying to benchmark ls gives me an unrecognized command error:
image
This makes sense because ls is not a command on Windows with cmd.exe.

Setting the shell manually with `hyperfine -S nu 'ls' --show-output` gives me another error:
image

nushell only supports running commands using `nu -c`.

FilipAndersson245:
Ensuring we only use cmd-specific command building when cmd is in use, as this prevents alternative shells to be run on windows, like Nushell.

Closes https://github.com/sharkdp/hyperfine/issues/568

### DIFF
--- a/src/benchmark/executor.rs
+++ b/src/benchmark/executor.rs
@@ -130,7 +130,13 @@ impl<'a> Executor for ShellExecutor<'a> {
     ) -> Result<(TimingResult, ExitStatus)> {
         let mut command_builder = self.shell.command();
         command_builder
-            .arg(if cfg!(windows) { "/C" } else { "-c" })
+            .arg(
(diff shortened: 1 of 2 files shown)


sharkdp:

I understand that this fixes the behavior for nushell, but how about other shells on Windows, like PowerShell?

FilipAndersson245:
I tested both and powershell supports both `/C` and `-c`, and both seem to work the same.

sharkdp:
No, looks good. Thank you!

PR #583, opened 2022-10-24
Default to $TERM unset = colour on Win32

nabijaczleweli:
This otherwise respects all other configuration (NO_COLOR, TERM=dumb, &c.), but it's very common for the environment to be lacking TERM, which in that case usually means a pro-verbial TERM=cmd.exe, which is colour-capable

Reported in https://github.com/sharkdp/hyperfine/commit/049d1138c94b1654a80b3afc164bcc5c672560f5#commitcomment-86934718


sharkdp:
Thank you!

PR #584, opened 2022-10-28
refactor: Update to clap 4

### LINKED ISSUE #578: Update to clap v4
see migration guide here: https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#migrating

berombau:
Closes #578 

- Followed the [migration guide](https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#migrating).
- Added features (help, usage, error-context) because of `default-features = false`.
- There could be some subtle breaking changes, but the existing tests were quite extensive.
- Removed `max_term_width(90)` in favor of wrap_help.
- The default new output of `--help` has some subtle changes, that all seem good, reducing the total output size from 161 lines to 133 lines (with equal width `COLUMNS=90`).
- The default template has changed things like color and capitalization. This is has been [discussed before with bat](https://github.com/sharkdp/bat/pull/2327).

TODO:
- There are some Builder patterns like usage strings e.g. `arg!` that could simplify things, but these are not used here.

clap v3 `hyperfine --help`
image

clap v4 `cargo run -- hyperfine --help`
image

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -122,7 +122,7 @@ impl<'a> Commands<'a> {
         let command_strings = matches.values_of("command").unwrap();
 
         if let Some(args) = matches.values_of("parameter-scan") {
-            let step_size = matches.value_of("parameter-step-size");
+            let step_size = matches.get_one::<String>("parameter-step-size").map(|s| s.as_str());
             Ok(Self(Self::get_parameter_scan_commands(
                 command_names,
                 command_strings,
--- a/src/command.rs
+++ b/src/command.rs
@@ -221,9 +221,8 @@ impl<'a> Commands<'a> {
                 return Err(OptionsError::TooManyCommandNames(command_strings.len()).into());
             }
 
-            let command_list = command_strings;
-            let mut commands = Vec::with_capacity(command_list.len());
-            for (i, s) in command_list.iter().enumerate() {
+            let mut commands = Vec::with_capacity(command_strings.len());
+            for (i, s) in command_strings.iter().enumerate() {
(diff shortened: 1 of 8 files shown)


sharkdp:

The "Min. supported rust version" check currently fails. We might need to bump the min. supported Rust version (in `.github/workflows/CICD.yml` and in the README).

> There could be some subtle breaking changes, but the existing tests were quite extensive.

Well... hopefully not. Do you have any suspicions? Maybe parts of your changes where you're not 100% sure that the migration did the right thing? If so, we should at least check manually. Or add more integration tests, ideally.

> There are some Builder patterns like usage strings e.g. `arg!` that could simplify things, but these are not used here.

Okay. I guess that could be done later?

src/command.rs
```diff
@@ -148,7 +154,7 @@ impl<'a> Commands<'a> {
                     bail!("Duplicate parameter names: {}", &duplicates.join(", "));
                 }
             }
-            let command_list = command_strings.collect::<Vec<&str>>();
+            let command_list = command_strings;
```
sharkdp, line 157: Hm. Do we need both? Or could we use `command_list` above?

berombau:
The MSRV of clap v4 is 1.60.0, so I bumped it to that version.
The subtle breaking changes are what the clap migration docs said, but haven't found any so far.

src/command.rs
berombau: I removed the unneeded command_list variables

sharkdp (approved):
Thank you!

PR #592, opened 2022-12-02
Update README.md

baarkerlounger:
Hyperfine is available in Debian testing and unstable repositories now which seems worth pointing to.


sharkdp:
Thank you

PR #603, opened 2023-01-07
Always sort exported results (fixes #601)

### LINKED ISSUE #601: Sort output markdown results by Relative column (ascending)
The first row would be 1.00 and then you grow from there. That'd make the results easier to interpret.

SUPERCILEX:


sharkdp:
Thank you for your request. I usually like to see the benchmarks in the order that I entered them. I'm not convinced that this should be the new default.

sharkdp:
Thank you. I think I'd like to go with a new CLI option. See #614. I'll close this PR, but feel free to tell me if you want to reopen it in case you want to work towards #614.

PR #604, opened 2023-01-07
Export user and system time

### LINKED ISSUE #600: Display parameter values if they aren't used in the benchmarked command
Not sure what the UI should look like, but if you only use parameter values in a prepare command, you currently have no way of knowing which benchmarks used which parameters.

### LINKED ISSUE #602: Include user and system time in exports
These are useful as a measure of efficiency when timings are similar.

SUPERCILEX:
Fixes #602


sharkdp:
closing for now, see #602

Issue #569, opened 2022-09-18
ARM64 Windows binaries


rashil2000:
Hi! 

I would like to use hyperfine on a Windows ARM server for running benchmarks. Since performance is important here, I think it will be better to run ARM64 natively rather than using x86_64 emulation. 

Ref: [starship](https://github.com/starship/starship/blob/master/.github/workflows/release.yml) provides ARM64 binaries as part of their releases. They are built using GitHub Actions.

sharkdp:
Is that the aarch64-pc-windows-msvc target? Sure, we could simply try to add that in the `.github/workflows/CICD.yml` file. Feel free to open a PR.

sharkdp:
I'm afraid it's not really possible to set this up in CI (see #621). You will probably have to compile natively on your host. See instructions here: https://github.com/sharkdp/hyperfine#with-cargo-linux-macos-windows



> Since performance is important here, I think it will be better to run ARM64 natively rather than using x86_64 emulation.

It's probably important that the application that you are benchmarking is a native one. But it might be okay to run hyperfine with x86_64 emulation.

PR #621, opened 2023-03-06
CI: build for aarch64-pc-windows

CarterLi:
Try fixing https://github.com/sharkdp/hyperfine/issues/569 which is also what I need

I know very little of rust. Hopefully it works


CarterLi:
CI failed because aarch64 binary could not be run in x86-64 platform, I have no idea how to fix it.

sharkdp:
> CI failed because aarch64 binary could not be run in x86-64 platform, I have no idea how to fix it.

This is not the only problem. It also doesn't really cross-compile:
```
[cross] warning: `cross` does not provide a Docker image for target aarch64-pc-windows-msvc, specify a custom image in `Cross.toml`.
[cross] note: Falling back to `cargo` on the host.
```

The problem is that `aarch64-pc-windows-msvc` is not a supported target for `cargo cross` (https://github.com/cross-rs/cross#supported-targets).

Also, I guess GA doesn't provide any native Windows runners on aarch64.

I'm afraid we have to close this.

Issue #615, opened 2023-02-19
Support `-` as shorthand for stdout for output


horenmar:
Related to my other issue, I tried using `--export-markdown -` to get the markdown output shown immediately in stdout. To my surprise, I ended up with a file named `-`.

Since using `-` for stdout/stderr is relatively common convention, I would like it to be supported for the output options as well.

sharkdp:
I use `--export-markdown /dev/stdout`, but supporting `-` sounds good. Thank you for the suggestion.

PR #623, opened 2023-03-07
PR for issue #615

humblepenguinn:
This PR adds the feature requested in issue #615

### DIFF
--- a/src/export/mod.rs
+++ b/src/export/mod.rs
@@ -1,3 +1,4 @@
+use std::env;
 use std::fs::{File, OpenOptions};
 use std::io::Write;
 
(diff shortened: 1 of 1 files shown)


src/export/mod.rs
```diff
@@ -89,9 +90,18 @@ impl ExportManager {
             ExportType::Markdown => Box::new(MarkdownExporter::default()),
             ExportType::Orgmode => Box::new(OrgmodeExporter::default()),
         };
+
         self.exporters.push(ExporterWithFilename {
             exporter,
-            filename: filename.to_string(),
+            filename: if filename == "-" {
+                if env::consts::OS == "windows" {
+                    "con:".to_string()
```
sharkdp, line 98: Thank you. Have you tested this on Windows and Linux? 

To distinguish between Windows and Linux, we use `cfg!(…)` elsewhere in hyperfines source code. The difference is that it is a compile-time switch. Your runtime switch might be favorable in certain (weird) scenarios, but I think I'd prefer to use `if cfg!(windows)` here instead of the `env::consts::OS` check.

humblepenguinn: Yeah, I tested this on my linux machine and as well as my windows one. It seemed to be working

sharkdp:
Thanks

sharkdp:
A file named '-' is still being created. I will fix this.

PR #633, opened 2023-03-28
Fix get_cmap MatplotlibDeprecationWarning

nicovank:
```
[~] ~/hyperfine/scripts/plot_whisker.py -o result.jpg results.json
~/hyperfine/scripts/plot_whisker.py:37: MatplotlibDeprecationWarning: The get_cmap function was deprecated in Matplotlib 3.7 and will be removed two minor releases later. Use ``matplotlib.colormaps[name]`` or ``matplotlib.colormaps.get_cmap(obj)`` instead.
  cmap = plt.cm.get_cmap("rainbow")
```

After this change, the warning disappears, and the output plot looks unchanged.

### DIFF
--- a/scripts/plot_whisker.py
+++ b/scripts/plot_whisker.py
@@ -34,7 +34,7 @@ else:
 times = [b["times"] for b in results]
 
 boxplot = plt.boxplot(times, vert=True, patch_artist=True)
-cmap = plt.colormaps["rainbow"]
+cmap = plt.get_cmap("rainbow")
 colors = [cmap(val / len(times)) for val in range(len(times))]
 
 for patch, color in zip(boxplot["boxes"], colors):
--- a/scripts/plot_whisker.py
+++ b/scripts/plot_whisker.py
@@ -34,7 +34,7 @@ else:
 times = [b["times"] for b in results]
 
 boxplot = plt.boxplot(times, vert=True, patch_artist=True)
-cmap = plt.cm.get_cmap("rainbow")
+cmap = plt.colormaps["rainbow"]
 colors = [cmap(val / len(times)) for val in range(len(times))]
 
 for patch, color in zip(boxplot["boxes"], colors):


sharkdp:
Thank you. Does this put a new upper limit on the used matplotlib version?

nicovank:
Matplotlib 3.7 [was just released](https://discourse.matplotlib.org/t/matplotlib-devel-ann-matplotlib-3-7-0/23568), but it looks like this was actually [deprecated since 3.6](https://matplotlib.org/3.6.0/api/cm_api.html#matplotlib.cm.get_cmap).

sharkdp:
Sorry, I meant: lower limit. In the sense of: this will probably not work on (very) old versions of matplotlib. How far back?

vadcx:
@sharkdp Yes, looks like the new method bumps the minimum version to 3.5:

> In Matplotlib 3.5 we added a .ColormapRegistry class and exposed an instance at the top level as matplotlib.colormaps. The existing top level functions in matplotlib.cm (get_cmap, register_cmap, unregister_cmap) were changed to be aliases around the same instance. In Matplotlib 3.6 we have marked those top level functions as pending deprecation. [source 3.7.0](https://github.com/matplotlib/matplotlib/blob/70e0ba8586e2d0bb38067154e077d920b218275b/doc/api/prev_api_changes/api_changes_3.7.0/deprecations.rst#deprecation-of-top-level-cmap-registration-and-access-functions-in-mplcm)

> Colormaps are now managed via matplotlib.colormaps (or .pyplot.colormaps), which is a .ColormapRegistry. While we are confident that the API is final, we formally mark it as experimental for 3.5
> [source 3.5.0](https://github.com/matplotlib/matplotlib/blob/c1c4b5a3aee1d1fd86ecb7fbc900c77af1034227/doc/users/prev_whats_new/whats_new_3.5.0.rst#colors-and-colormaps)

> For 3.9 (or later): Remove all the functions we want to remove. matplotlib/matplotlib#20853

nicovank:
Correct. I found that out last week but didn’t have time to update this PR. I think this can stay as-is at least until `get_cmap` is officially removed. Maybe there’s a way to remove usage entirely, I’ll try to take another look today.

vadcx:
@nicovank Literally, why not both? The simplest imho is to dynamically choose either of them and call the existing variant (as long the underlying API is the same? I didn't check). That's gonna be the fastest solution?

If you want to recreate the rainbow colors it's as simple as iterating over the HSL color space: different degree 0-360° on the spectrum with maximum saturation and luminance.

If you want to exercise in the above, there's the human-friendlier version https://www.hsluv.org/ :) The HSL iteration is a couple of lines of code in any language, HSLuv is going to be a required dependency for a minimal effect.

nicovank:
As per https://github.com/matplotlib/matplotlib/issues/20853:
> Keep `plt.get_cmap()` - This is too widely used to be removed.

So I think it's best here to just use that (versus `plt.cm.get_cmap`).

AFAIK this does not set a new lower bound to the used Matplotlib. I tested on my machine down to 2.0.0 (released January 2017). Beyond this, it fails for some unrelated (`tkinter`) reason.

sharkdp:
> I tested on my machine down to 2.0.0 (released January 2017).

Okay. So the following statement was incorrect?

> looks like the new method bumps the minimum version to 3.5:

nicovank:
It was accurate with the previous change (see commit history). `plt.get_cmap` (current version) seems to have been supported for a long time.

sharkdp:
> It was accurate with the previous change (see commit history). `plt.get_cmap` (current version) seems to have been supported for a long time.

Oh, sorry. I missed the new commit.

Thank you!

Issue #639, opened 2023-04-17
Replace usage of 'winapi' with 'windows-sys'


sharkdp:
See https://github.com/sharkdp/hyperfine/issues/624#issuecomment-1508975294

PR #641, opened 2023-04-17
Use windows-sys and fix feature flag

clemenswasser:
Implements #639
Also fixes the `windows_process_extensions_main_thread_handle` feature toggle so that It now actually works 😄

### DIFF
--- a/src/timer/mod.rs
+++ b/src/timer/mod.rs
@@ -13,6 +13,9 @@ use std::fs::File;
 #[cfg(target_os = "linux")]
 use std::os::unix::io::AsRawFd;
 
+#[cfg(target_os = "windows")]
+use windows_sys::Win32::System::Threading::CREATE_SUSPENDED;
+
 use crate::util::units::Second;
 use wall_clock_timer::WallClockTimer;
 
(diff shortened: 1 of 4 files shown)


src/timer/mod.rs
```diff
@@ -83,7 +86,7 @@ pub fn execute_and_measure(mut command: Command) -> Result<TimerResult> {
         use std::os::windows::process::CommandExt;
 
         // Create a suspended process
-        command.creation_flags(4);
+        command.creation_flags(CREATE_SUSPENDED);
```
sharkdp, line 89: :+1: 

Maybe we can remove the comment now? (Or explain what a suspended process is?)

src/timer/windows_timer.rs
```diff
@@ -48,24 +52,24 @@ pub struct CPUTimer {
...
         let job_object = unsafe { CreateJobObjectW(ptr::null_mut(), ptr::null_mut()) };
-        assert!(!job_object.is_null(), "CreateJobObjectW failed");
+        assert!(job_object != 0, "CreateJobObjectW failed");
 
         // SAFETY: The job object handle is valid
         let ret = unsafe { AssignProcessToJobObject(job_object, child_handle) };
         assert!(ret != 0, "AssignProcessToJobObject failed");
 
-        #[cfg(windows_process_extensions_main_thread_handle)]
+        #[cfg(feature = "windows_process_extensions_main_thread_handle")]
```
sharkdp, line 65: If I understand/recall correctly, the intention was that this version could be used for nightly-compiler builds (with `windows_process_extensions_main_thread_handle` additionally activated), and once that feature is stabilized, we can remove the alternative `not(feature = …)` version below and make this the main code path.

sharkdp (approved):
Very nice. A clear improvement.

src/timer/mod.rs
clemenswasser: I hope the explanation helps.

src/timer/windows_timer.rs
clemenswasser: Yes, this is the plan 👍

PR #655, opened 2023-06-03
Add new --sort option

### LINKED ISSUE #601: Sort output markdown results by Relative column (ascending)

### LINKED ISSUE #614: Allow picking sort order of the output
When I run `hyperfine --parameter-scan version 0 9 './some-command-{version}`, I get the results sorted by their speed. This can be often useful, which is why I believe it became the default, but I am specifically iterating over 10 different versions to get time series data, so I want the output in order.

I did this by asking for markdown output, which is in-order, but people also request that to be sorted by speed (#603, #601). Meanwhile, I want the default output sorted by command order...

I think the solution is to make the sort order customizable for all outputs.

sharkdp:
This adds a new `--sort` option to choose the way in which the results in the speed comparison and the markup exports are ordered.

Default behavior, `--sort auto` or `--sort mean-time`:
image

With `--sort command`
image

Markdown export with `--sort mean-time`:

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `sleep 0.1` | 101.1 ± 0.5 | 100.6 | 103.1 | 1.00 |
| `sleep 0.2` | 200.9 ± 0.3 | 200.7 | 201.9 | 1.99 ± 0.01 |
| `sleep 0.3` | 301.0 ± 0.2 | 300.8 | 301.3 | 2.98 ± 0.01 |

closes #614
closes #601

### DIFF
--- a/src/benchmark/relative_speed.rs
+++ b/src/benchmark/relative_speed.rs
@@ -1,7 +1,7 @@
 use std::cmp::Ordering;
 
 use super::benchmark_result::BenchmarkResult;
-use crate::util::units::Scalar;
+use crate::{options::SortOrder, util::units::Scalar};
 
 #[derive(Debug)]
 pub struct BenchmarkResultWithRelativeSpeed<'a> {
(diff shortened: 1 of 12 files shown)


SUPERCILEX:
This is awesome, thanks!

Issue #507, opened 2022-05-09
Problems with the default color style in Windows Powershell


Pietruszek:
After https://github.com/sharkdp/hyperfine/issues/427 has been resolved there's now a default coloring scheme enabled for Windows. 

And while it looks good in the `cmd` or the default color scheme for the Windows Terminal, in Powershell the background is blue. What that means for the `hyperfine`'s color scheme is that the below output:

```
PS D:\Users\ZPO\Documents\benchmarksgame\nbody\Picked> hyperfine -N --warmup 2 './Cs_projects/nbody_cs/bin/x64/Release/netcoreapp3.1/nbody_cs.exe 50000000' './nbody_gpp.exe 50000000' 'java nbody 50000000'
Benchmark 1: ./Cs_projects/nbody_cs/bin/x64/Release/netcoreapp3.1/nbody_cs.exe 50000000
  Time (mean ± σ):      7.086 s ±  0.029 s    [User: 7.062 s, System: 0.034 s]
  Range (min … max):    7.055 s …  7.145 s    10 runs

Benchmark 2: ./nbody_gpp.exe 50000000
  Time (mean ± σ):      6.948 s ±  0.019 s    [User: 6.939 s, System: 0.000 s]
  Range (min … max):    6.925 s …  6.976 s    10 runs

Benchmark 3: java nbody 50000000
  Time (mean ± σ):      7.198 s ±  0.025 s    [User: 7.175 s, System: 0.058 s]
  Range (min … max):    7.165 s …  7.249 s    10 runs

Summary
  './nbody_gpp.exe 50000000' ran
    1.02 ± 0.00 times faster than './Cs_projects/nbody_cs/bin/x64/Release/netcoreapp3.1/nbody_cs.exe 50000000'
    1.04 ± 0.00 times faster than 'java nbody 50000000'
```

displays like that:

![obraz](https://user-images.githubusercontent.com/32390629/167421656-ef2689b0-b9eb-47e1-893c-3c08c0c05ced.png)

so some info (like the User and System times) is very hard to read and some (like the max times or the strings after "faster than") is exactly the same color as the background.

This issue can be mitigated either by disabling the coloring using the `--style` option or by using a different shell (Windows Terminal or `cmd`), but considering that some users might be testing `hyperfine` for their first time in regular Powershell it is quite bad UX-wise.

Though I'm not really sure what can be done to fix this...

Two attempts at solving this I could imagine are:
[...]

PR #661, opened 2023-06-23
Use only colors supported by conhost.exe

Gordon01:
The previous terminal for Windows is `conhost.exe` (the one with blue background) which renders some colors differently. 
Green and Red, for example, are quite dim, whereas, on typical unix terminal and modern Windows Terminal, they're indistinguishable from "normal" ones.

This patch changes the overall `hyperfine` color scheme to a few "safe" colors in the following manner.

<details><summary>Color scheme</summary>
<p>

#### Main measurement text:
    green -> yellow bold
    green bold -> yellow bold

#### Min:
    cyan -> bright green

#### Max:
    purple -> bright red

#### Other:
    blue -> cyan
    magenta -> cyan

</p>
</details> 

## Before:

<details><summary>conhost.exe</summary>
<p>

image
image

</p>
</details> 

<details><summary>Windows Terminal</summary>
<p>

image

</p>
</details> 

<details><summary>VS Code</summary>
<p>

image

</p>
</details> 

## After:

<details><summary>conhost.exe</summary>
<p>

image
image

</p>
</details> 

<details><summary>Windows Terminal</summary>
<p>

image

</p>
</details> 

<details><summary>VS Code</summary>
<p>

image

</p>
</details> 

This PR addresses https://github.com/sharkdp/hyperfine/issues/507

### DIFF
--- a/src/benchmark/mod.rs
+++ b/src/benchmark/mod.rs
@@ -306,7 +306,7 @@ impl<'a> Benchmark<'a> {
 
                 println!(
                     "  Range ({} … {}):   {:>8} … {:>8}    {}",
-                    "min".bright_yellow().bold(),
+                    "min".bright_green(),
                     "max".bright_red(),
                     min_str.bright_green(),
                     max_str.bright_red(),
--- a/src/benchmark/mod.rs
+++ b/src/benchmark/mod.rs
@@ -231,7 +231,7 @@ impl<'a> Benchmark<'a> {
 
             let msg = {
                 let mean = format_duration(mean(&times_real), self.options.time_unit);
-                format!("Current estimate: {}", mean.to_string().green())
+                format!("Current estimate: {}", mean.to_string().yellow().bold())
             };
 
             if let Some(bar) = progress_bar.as_ref() {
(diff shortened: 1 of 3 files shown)


Gordon01:
Sorry forgot to change `min` label color. Now it's correct:

image

sharkdp:
Thank you. Maybe it's just me, but I liked the previous colors better. Is there any way we can do a minimal fix without changing everything? Possibly only on Windows, or ideally only on "affected" terminal emulators?

Gordon01:
I guess it's pretty hard to reason about what colors "are better". I'm definitely not ready for it...

> Possibly only on Windows, or ideally only on "affected" terminal emulators?

It's doable, I can try if you're fine with additional code

sharkdp:
I would favor a solution where we put all of the used colors in one central place. And then we could have two variants of the "theme", one for cases where we assume that we are running in a old-style blue Windows terminal, and the default theme.

sharkdp:
I'm closing this due to inactivity. Please feel free to comment in case it should be re-opened.

Issue #607, opened 2023-01-28
Add an option to save a benchmark run for later comparisons


CosmicHorrorDev:
There are times when I want to have some baseline saved that I compare later changes to. It would be convenient if I had a way to save a benchmarking run and then load that to later runs. Something like

```text
$ hyperfine --save-benchmark baseline 'sleep 1'
$ hyperfine --load-benchmark 'sleep 0.5'
{{ display things as though `hyperfine 'sleep 1' 'sleep 0.5'` had been run }}
```

If you're interested in supporting this then I can work on adding it as a feature

sharkdp:
That sounds like a cool idea, thank you. It also sounds like a rather complex feature that is not easy to design. So before we go ahead trying to implement this, I'd like to discuss how this would work, what kind of implications this has on other features, what the CLI would look like, etc.

Note: we already have `--export-json`, so that could probably be used for the "storing" side of things.

sharkdp:
*Slightly* related: #577

miluoshi:
I would also welcome this feature. It's currently possible to save report with `--export-json` so could it be implemented in a way, that instead of defining one command, there would be a `--import-json` flag?

```sh
# before - a single run
hyperfine 'command1' 'command2'

# after - two separate runs
hyperfine 'command1' --export-json baseline.json
hyperfine --import-json baseline.json 'command2'
```

The position of arguments would determine the order of compared commands the same way as position of commands in the 1st example. What do you think?

RalfJung:
In Miri we are planning to hand-implement an ad-hoc approximation of this since we can't really use the existing comparison support (https://github.com/rust-lang/miri/issues/3999). Would be amazing to get proper support for this upstream. :)

Walther:
Adding a broader thought: "have you tried rubbing a database on it?"[^1] Could it be worthwhile for `hyperfine` to add support for slightly more formal, database-shaped data processing and output?

The current `json` output is a nice and useful feature. I've used it extensively for experiments conducted for my master's thesis. Being able to collect all the data into a single file, across a wide variety of parameters provided via `--parameter-list`, including combinatorics with multiple lists, is absolutely wonderful. However, analyzing the results is a bit tough. My post processing scripts are fairly involved, in order to transform the data into a more table-shaped, query-able form. Nushell has proved to be handy here, but it isn't ideal.

Currently, the in-memory presentation is the `BenchmarkResult` struct, and serialization is implemented fairly manually for each output format under the `export` directory.

[...]

PR #671, opened 2023-07-29
Load previous runs

richardmon:
This is an initial attempt to solve this feature request https://github.com/sharkdp/hyperfine/issues/607 and https://github.com/sharkdp/hyperfine/issues/577, it takes a exported json and imports it using the command option ``--import-json output.json``.

I'm open to suggestions :^)


sharkdp:
Thank you very much for your contribution. This looks very interesting. Could you please go into a bit more detail on how to use this? Can I use `--import-json` without adding any new benchmarks? Can I add as many new benchmarks as I want? Can I use `--import-json` multiple times? Does this clash with any of the existing options (e.g. parameter-scan/list, export)?

sharkdp:
I'm closing this due to inactivity. Please feel free to comment in case it should be re-opened.

PR #675, opened 2023-08-17
Change Winget Releaser job to `ubuntu-latest`

sitiom:
See https://github.com/sharkdp/fd/pull/1364:
> Winget Releaser now supports non-Windows runners, and `ubuntu-latest` is generally faster.


sharkdp:
Thank you

PR #693, opened 2023-10-16
Remove dependency on `atty` and use `std::io::IsTerminal` instead

acuteenvy:
Since the [MSRV](https://github.com/sharkdp/hyperfine/commit/d91c126a9be8ea8d96c219d44ac8f458e402ffa1) is now 1.70, the implementation from `std` can be used instead.


sharkdp:
Thank you!

PR #706, opened 2023-11-20
Add `-E` and `--show-elapsed` flag to show the elapsed time.

jooris-hadeler:
# Changes
- add  the `-E` and `--show-elapsed` flags to optionally enable a elapsed time counter

# Origin
Recently when I was benchmarking software for my university course, I found that `hyperfine` gives you very little information about how long the current run has been running. This is very unfortunate because you can't tell if your program has possibly deadlocked. This pull request adds the `-E` and `--show-elapsed` flags which will show you for how long the current run is running.

### DIFF
--- a/src/benchmark/executor.rs
+++ b/src/benchmark/executor.rs
@@ -215,7 +215,8 @@ impl<'a> Executor for ShellExecutor<'a> {
             }
 
             if let Some(bar) = progress_bar.as_ref() {
-                bar.inc(1)
+                bar.inc(1);
+                bar.reset_elapsed();
             }
         }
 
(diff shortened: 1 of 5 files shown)


jooris-hadeler:
I must say I completely forgot to check if someone had already proposed this. Compared to #416 it is quite similar, although in comparison to #581 this would always show the time for the current run, regardless of it being the initial or a followup run. I think this feature is only really necessary in some specific cases (e.g. the mentioned long runs where it might crash), I think it would be wise to hide this behind a flag.

jooris-hadeler:
It has been two years since I opened this Pull request. The last communication on this thread was that there is already someone who implemented this in a PR, yet after two years none of these have been merged. I would love to know if we can get this PR merged or if I should close this PR?

sharkdp:
Let's please leave it open, I am still planning to do a hyperfine 2.0 release soon (I'm not giving any specific dates), and I am planning to integrate this as well.

Sorry that it takes so long.

jooris-hadeler:
Okay no worries, if there is anything you want changed I'm happy to integrate your feedback.

PR #717, opened 2024-01-05
Add clap::ValueHint to fix zsh completion

xzfc:
Problem description: in zsh, type `hyperfine ` (with a space) and press tab. As a result, it completes nothing, not even the file names (the default fallback if no completion function is found), which makes the generated completion not very useful.

This PR adds value hints to make generated completion behave properly.


sharkdp:
Thank you very much.

PR #719, opened 2024-01-21
Add conclude cli argument

### LINKED ISSUE #565: Conclude phase (opposite of --prepare)
In the same manner as the --prepare phase exists, the feature request is for a non-timed --conclude or --post phase that would run after each benchmarked command. There would each be 1 or N number of these.

In my use case, I cannot do the cleanup after the command executes W(armup)+N times. I also understand that running in this phase can pollute the cache, but in my case, the benchmarks can already run for hours.

The current workaround is to put the per command shutdown steps in each bechmarked command but that pollutes the benchmarking times.

Thanks again for an excellent utility

jackoconnordev:
## Description

Resolves #565 - "Conclude phase (opposite of --prepare)"

This is an implementation of `--conclude` argument where `--conclude` is implemented in the same manner as `--prepare.`

**Example usage**
`hyperfine --warmup 1 --runs 2 --prepare 'echo prepare >> test.txt' --conclude 'echo conclude >> test.txt' 'sleep 0.5'`

_Output_
```
prepare
conclude
prepare
conclude
prepare
conclude
```

Relevant tests have been added and relevant docs updated.

## Discussion

I'm very open to PR feedback and I don't mean to appear out of nowhere and force this change on anyone. I do think the --conclude option makes a lot of sense. I found the codebase for this project quite approachable after coming across it in https://nnethercote.github.io/perf-book/benchmarking.html and I thought attempting this PR would be a good educational exercise.

This was quite straight-forward to implement using existing --prepare code as a template and I naturally ended up with a solution basically identical to what @shoffmeister had linked in a comment in the issue thread.
Despite prepare and conclude implementions being almost identical (and setup and cleanup not far off either) I'm not sure about the payoff for coming up with some abstraction which increases code reuse between them.

I can't imagine there ever being a need to introduce another execution scheduling option beyond `setup/prepare/conclude/cleanup`. The Cypress test runner in javascript land only has BeforeAll, Before, After, AfterAll for running extra code around tests. The Cypress naming scheme could be support for future aliases for this functionality which are of the form `before-*,after-*`.

Any major benefit from additional work I think would be in maintainability and the current code seems readable enough to me as is (although that's subjective).
I did see this is in the hyperfine 2.0 milestone list though so it's possible there are already plans for this issue which are such a rework and fix problems I'm just not thinking of since I'm not familiar enough.

## New tests

- conclude_commands_are_executed_after_each_timing_run
- conclude_commands_are_executed_after_each_warmup
- setup_prepare_conclude_cleanup_combined
- fails_with_wrong_number
- fails_for_unknown_conclude_command
- takes_conclusion_command_into_account_for_computing_number_of_runs
- takes_both_preparation_and_conclusion_command_into_account_for_computing_number_of_runs

## Example Command Output

Below is example output of running the release build of hyperfine build from this PR with setup/prepare/conclude/cleanup arguments all being passed.

```
   Compiling hyperfine v1.18.0 (/home/jackoconnordev/Code/open-source/hyperfine)
    Finished release [optimized] target(s) in 10.25s
     Running `target/release/hyperfine -L n 1,2 -r 2 --show-output --setup 'echo setup n={n}' --prepare 'echo prepare={n}' --conclude 'echo conclude={n}' --cleanup 'echo cleanup n={n}' 'echo command n={n}'`
Benchmark 1: echo command n=1
setup n=1
prepare=1
command n=1
conclude=1
prepare=1
command n=1
conclude=1
  Time (mean ± σ):       0.0 µs ±   0.0 µs    [User: 0.0 µs, System: 60.2 µs]
  Range (min … max):     0.0 µs …   0.0 µs    2 runs
 
  Warning: Command took less than 5 ms to complete. Note that the results might be inaccurate because hyperfine can not calibrate the shell startup time much more precise than this limit. You can try to use the `-N`/`--shell=none` option to disable the shell completely.
 
cleanup n=1
Benchmark 2: echo command n=2
setup n=2
prepare=2
command n=2
conclude=2
prepare=2
command n=2
conclude=2
  Time (mean ± σ):      90.0 µs ± 127.2 µs    [User: 320.5 µs, System: 167.7 µs]
  Range (min … max):     0.0 µs … 179.9 µs    2 runs
 
  Warning: Command took less than 5 ms to complete. Note that the results might be inaccurate because hyperfine can not calibrate the shell startup time much more precise than this limit. You can try to use the `-N`/`--shell=none` option to disable the shell completely.
 
cleanup n=2
Note: The benchmark comparison could not be computed as some benchmark times are zero. This could be caused by background interference during the initial calibration phase of hyperfine, in combination with very fast commands (faster than a few milliseconds). Try to re-run the benchmark on a quiet system. If you did not do so already, try the --shell=none/-N option. If it does not help either, you command is most likely too fast to be accurately benchmarked by hyperfine.
```


sharkdp:
Thank you for the excellent PR. I plan to take a look soon.

sharkdp (approved):
Thank you!

jackoconnordev:
Happy to help 😁

PR #727, opened 2024-03-02
Nicer whiskers plot

serpent7776:
- add missing encoding when opening input file
- add labels to x ticks, rotated by 45 deg
- sort data by median, ascending

I like this layout much more, but if needed I can hide new functionality behind command line flags.

### DIFF
--- a/scripts/plot_whisker.py
+++ b/scripts/plot_whisker.py
@@ -15,6 +15,7 @@ import matplotlib.pyplot as plt
 parser = argparse.ArgumentParser(description=__doc__)
 parser.add_argument("file", help="JSON file with benchmark results")
 parser.add_argument("--title", help="Plot Title")
+parser.add_argument("--sort-by", choices=['median'], help="Sort method")
 parser.add_argument(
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you!

Can you provide a before-and-after screenshot?

> sort data by median, ascending

This would be nice to have as an option, but I prefer benchmarks to be ordered by input ordering — by default.

serpent7776:
Sure, here's the old version:

plot_old

This is the new version (one issue is that longer labels are cut off):

plot_new

I can fix the issue by applying `plt.figure(figsize=(10, 6), constrained_layout=True)`, but I don't know why this works:

plot_sized

sharkdp:
Cool — looks great! If we can hide the sorting feature behind a command line option (maybe `--sort-by=…`), I'm happy to merge this.

serpent7776:
Updated the code.

sharkdp (approved):

sharkdp:
Thank you

PR #728, opened 2024-03-08
Build binaries for aarch64-apple-darwin

Phault:
This uses the macOS 14 Arm64 [beta] runner to build binaries for Apple silicon.


sharkdp:

PR #739, opened 2024-04-04
Fix the Arch url to repo in README.md

Aym3nJ:
PS: I apology in advance if i messed something up , i tried to read Contribution.md but couldn't find one!


sharkdp:
Thank you!

PR #753, opened 2024-06-23
Initial work on tracking memory usage

sharkdp:


xHyroM:
excited to see this feature :), can be linked with https://github.com/sharkdp/hyperfine/issues/86

sharkdp:
There's no harm in merging this as is. It seems to work fine (on Linux) and allows me to further design changes for Hyperfine 2.0

sharkdp:
> It seems to work fine

Not really, actually. We need to use `wait4` instead of `getrusage` to make this work properly.

PR #758, opened 2024-07-10
add: legend modification parameters and output DPI

Spreadcat:
This commit adds a CLI parameter to plot_histogram in order to position the legend of the plot on the diagram.
It also defines the output DPI for the plot to 600 dpi.

### DIFF
--- a/scripts/plot_histogram.py
+++ b/scripts/plot_histogram.py
@@ -14,7 +14,7 @@ parser.add_argument("--title", help="Plot title")
 parser.add_argument(
     "--labels", help="Comma-separated list of entries for the plot legend"
 )
-parser.add_argument( "--bins", help="Number of bins (default: auto)")
+parser.add_argument("--bins", help="Number of bins (default: auto)")
 parser.add_argument(
     "--lloc", help="Location of the legend on plot (default: upper center)",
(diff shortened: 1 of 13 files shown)


Spreadcat (commented):
cleaned up in the changes.

scripts/plot_histogram.py
```diff
@@ -14,6 +15,11 @@
     "--labels", help="Comma-separated list of entries for the plot legend"
 )
 parser.add_argument("--bins", help="Number of bins (default: auto)")
+parser.add_argument(
+    "--lloc", help="Location of the legend on plot (default: upper center)",
```
sharkdp, line 19: "lloc" doesn't seem to be a standard abbreviation, not even in matplotlib, which is full of cryptic terms. Would it hurt to call this `--legend-location`?

Spreadcat: No problem there. It was just lazy-typing to use "lloc" instead of "legend-location". I've change the name of the parameter as you can see. But wait with the merge. I've just quickly wanted to test the script to make sure the changes work and I got a python error. I'll just have a look at this and then put the MR back to _active/no-draft_ when ready.

sharkdp:

PR #768, opened 2024-09-15
Add completions for Python scripts

EmilyGraceSeville7cf:
# Summary

Add completions for Python scripts:

image

# Features

- [x] Completions for options
- [x] Completions for option values (default values are emphasized)

# Requirements

- `>= fish 3.7.1`

### DIFF
--- a/completions/advanced_statistics.fish
+++ b/completions/advanced_statistics.fish
@@ -1,0 +1,6 @@
+set command advanced_statistics.py
+
+complete -c $command \
+    -s h \
+    -l help \
+    -d 'Show help'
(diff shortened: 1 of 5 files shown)


sharkdp:
Thank you. This is certainly nice for fish users, but to be honest, I don't want to have to maintain completion scripts for all kinds of shells (unless they are auto-generated somehow). For `hyperfine` itself, sure. But for those Python scripts, it seems too much work for too little benefit (we would also have to package & install these properly, and advertise them in the documentation)  — sorry.

EmilyGraceSeville7cf:
I don't know Rust for now (at least yet) so I can't write a code for their automatic generation. Although I am able to support them. However, I see ur point- I can move these completions to Fish shell itself.

sharkdp:
> I can move these completions for Fish shell itself.

That would be great; thank you

PR #775, opened 2024-11-11
Multiple output policies

### LINKED ISSUE #529: `--output` overwrites previous commands and runs
Currently I'm working on a set of benchmarks where I want to save the output of (at least) one run from each. `--output` should arguably be allowed N times, where N is the number of commands being benchmarked. 

However, I would also argue that keeping each runs output is extremely useful, specifically for debugging/figuring out why statistical outliers exist.

As an arbitrary example:
```
$ hyperfine --output ./example.log "ls" "pwd"
```
will only show output from a single `pwd`. Passing multiple outputs does not work, as hyperfine only expects one.

As a proposal for how this could change there are two things that would be nice to see:
1. Allow multiple outputs, the same way `--prepare` is allowed per command
   ```
   $ hyperfine --output ./ls.log "ls" --output ./pwd.log "pwd"
   ```
2. Allow some parameter, like N, for the run number
   ```
   $ hyperfine --output "./ls-{N}.log" "ls"
   ```

For short running commands like `ls` this is a problem as it may run somewhere in the neighborhood of 5k times, so it may be a problem to create a new logfile for each run. Longer running commands obviously do not have this problem.

sharkdp:
- Allow `--output` to appear *exactly once* or *exactly N times*, enabling several new use cases like `hyperfine --output=cmd1.log cmd1 --output=cmd2.log cmd2` or `hyperfine --output=null cmd --output=./file cmd`.
- In the `--help` text for `--output`, mention the `HYPERFINE_ITERATION` environment variable for users who want to log each and every iteration.

closes #529

### DIFF
--- a/doc/hyperfine.1
+++ b/doc/hyperfine.1
@@ -290,6 +290,11 @@ Don't redirect the output at all (same as \&'\-\-show\-output').
 .IP "<FILE>"
 Write the output to the given file.
 .RE
+.IP
+This option can be specified once for all commands or multiple times,
+once for each command. Note: If you want to log the output of each and
+every iteration, you can use a shell redirection and the $HYPERFINE_ITERATION
+environment variable: 'my-command > output-${HYPERFINE_ITERATION}.log'
 .HP
 \fB\-\-input\fR \fIWHERE\fP
 .IP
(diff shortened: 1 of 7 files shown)


StripedMonkey:
Thank you for your hard work. It's much appreciated!

sharkdp:
Thanks for patiently waiting two years for this :smile:

PR #806, opened 2025-03-05
Add plot_benchmarks scripts

marxin:
The script is intended for a collection of hyperfine JSON files where multiple benchmarks were run for a selected group of commands.

There are various projects that use such graphs:

https://github.com/rui314/mold?tab=readme-ov-file#mold-a-modern-linker
https://github.com/gimli-rs/addr2line?tab=readme-ov-file#performance

A sample output graph from the suggested script:
linker-benchmark

### DIFF
--- a/scripts/plot_benchmarks.py
+++ b/scripts/plot_benchmarks.py
@@ -1,0 +1,80 @@
+#!/usr/bin/env python
+# /// script
+# requires-python = ">=3.10"
+# dependencies = [
+#     "matplotlib",
+#     "pyqt6",
+#     "numpy",
+# ]
+# ///
+
+# This program shows `hyperfine` benchmark results as a bar plot grouped by command.
+
+import argparse
+import json
+import pathlib
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+parser = argparse.ArgumentParser(description=__doc__)
+parser.add_argument(
+    "files", nargs="+", type=pathlib.Path, help="JSON files with benchmark results"
+)
+parser.add_argument("--title", help="Plot Title")
+parser.add_argument(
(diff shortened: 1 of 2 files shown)


scripts/plot_benchmarks.py
```diff
@@ -0,0 +1,80 @@
...
+# This program shows `hyperfine` benchmark results as a bar plot grouped by command.
+
+import argparse
+import json
+import pathlib
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+parser = argparse.ArgumentParser(description=__doc__)
```
sharkdp, line 20: Should the command above be a docstring in order for this to work?

scripts/plot_benchmarks.py
```diff
@@ -0,0 +1,80 @@
...
+parser.add_argument(
+    "files", nargs="+", type=pathlib.Path, help="JSON files with benchmark results"
+)
```
sharkdp, line 23: Maybe we could add a note here that all files must include results with the same commands? Or maybe that should be described in the top-level doc string?

sharkdp (approved):
This looks great, thank you very much. Can we maybe find a more descriptive name for this script? Maybe something like `plot_benchmark_comparison` or `plot_grouped_comparison`?

marxin:
Thanks for the review. I selected one of the suggested script names.

sharkdp:
Thank you!

webpro:
Just added an update to #786 and saw this PR. Is it maybe interesting to support this use case so one can drop multiple JSON files onto the chart and perhaps even support URLs to achieve roughly the same (but without the need to script anything)? Venz also has a "download img as svg/png" option. Feel free to let me know, happy to consider more use cases if there's demand.

PR #808, opened 2025-03-15
Add option to name reference command

niklasdewally:
Add --reference-name option to give a reference command a name.

This was discussed in the PR that added --reference, #744, but was unfortunately never implemented.


sharkdp:
Thank you very much!

PR #809, opened 2025-03-21
Add interactive HTML export format

lukaskoebe:
Hi, I know this is quite a large PR, so if you feel like this kind of thing doesn't fit into this project no hard feelings. Just try it out and see if you like it.

My motivation was that I always loved the python ploting scripts for hyperfine, but I thought it would be great to have a standalone version of those plots that doesnt depend on having the python libs installed etc. and that can be created directly using hyperfine.

This PR implements a new HTML exporter that generates a standalone, interactive report with multiple visualizations:
- Summary table with relative performance metrics
- Boxplots, histograms and time progression charts
- Advanced statistics and parameter analysis

The report uses embedded JavaScript (Plotly.js) for interactive visualizations


sharkdp:
It might be a few days before I can try & review this, but I already love it! That is a great idea.

Issue #813, opened 2025-04-27
Allow getting parameters from file lines


Scripter17:
To benchmark a URL cleaning tool I'm building, I got 9000 URLs from reddit using `reddit.com/domain/$DOMAIN` and want to pass them all into a bash script that's effectively a fancy wrapper around hyperfine. The problem is that passing the first 3000 URLs into a `--parameter-list` errors on the argument list being too long, and passing in all 9000 somehow crashes my laptop.

What I'd like is to be able to write the URLs to a file then say, for example, `--parameters-from-file-lines url urls.txt`. In fact this could skip the file entirely by using [`--parameters-from-file-lines url <(url-getting-command)`](https://www.gnu.org/software/bash/manual/bash.html#Process-Substitution).

Scripter17:
**Update**: Since Sharkdp, the owner of Hypefine, seems to be no longer maintaining it, I've decided that the bounty will be paid out in full for each of the existing PRs assuming they work as needed when I test them and Sharkdp doesn't merge one before August 18th.

Any PRs submitted past this point (with the exception of wcj007 since they got here before this change) will not be eligible for the bounty unless it gets merged before that deadline.

~~I don't personally need this anymore but it's a useful enough feature that, if I'm allowed to, I'd like to put a $10 USD bounty on it. Preferably over paypal but I'm flexible.~~

~~The bounty will be paid once the PR is merged. If anything goes wrong after that you don't need to worry about refunding me.~~

The minimum requirements for the bounty are:

- A syntax like `--parameter-file name path/to/file.txt` that's basically equivalent to `--parameter-list name line1,line2,line3`.
- Accepting both \n and \r\n line separators. Optionally with  a mode that uses NUL instead like `grep -z`, but that probably steps on a few too many toes.
- Reading each file line by line. If needed an initial pass to count the lines is perfectly fine but what's important is not loading an entire 2 million line file into RAM all at once. If one single line is 3GB then that's on the user for giving Hyperfine stupid inputs.

If possible I'd like it to accept device files like Bash's `<(command)` but that raises a ton of weird issues and would probably ruin any benchmarks it's used in anyway, so I'm fine with it being either UB or detected and rejected.

If this ends up being a pain in the ass then feel free to request a $5 USD "Sorry for putting you through that" bonus. No questions asked; Just the honor system.

wcj007:
@sharkdp @Scripter17, before I start, could you please confirm the preferred contribution path given that PR #835 is still open?

[...]

PR #835, opened 2025-10-17
Implement reading parameters from files

fraserli:
This PR adds the `--parameter-list-file` option as described in #813. The option can also be provided in combination with `--parameter-list` while maintaining the correct ordering.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -201,6 +201,21 @@ fn build_command() -> Command {
                 ),
         )
         .arg(
+            Arg::new("parameter-list-file")
+                .long("parameter-list-file")
+                .action(ArgAction::Append)
+                .allow_hyphen_values(true)
+                .value_names(["VAR", "FILE"])
+                .conflicts_with_all(["parameter-scan", "parameter-step-size"])
+                .help(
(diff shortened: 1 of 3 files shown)


Scripter17:
@fraserli Hi. Not sure if you were notified by my message in the original feature request, but I recently offered a $10 USD bounty (+$5 USD if you personally found the change a pain in the ass to do) for this feature.

Shortly afterwards LeonDong0807 made a PR and after a few days I decided to check Hyperfine's activity and found its past 3 commits were 3, 5, and 8 months ago. As such I've changed the terms to be that if I find the PR to be good enough and Sharkdp doesn't respond by August 18th I'll pay out the bounty anyway.

Technically, under those terms you are also eligible. So I'll check out this PR in the next few days and, if it's also good and Sharkdp doesn't respond in time, I'll try to send you $10 as well.

I'd prefer to do this via paypal but I'm flexible. Just let me know.

Scripter17:
Hi again! Finally got around to testing this. My only real issue is that the code loads the entire file into memory at once, which for very large input files is... not good.

Fortunately one of the other PRs seems to fit my needs perfectly so it's no issue if you don't want to work on this again, but if you ever do there's $10 waiting for you.

PR #857, opened 2026-03-19
fix: expose HYPERFINE_ITERATION to --prepare and --conclude commands

### LINKED ISSUE #781: ${HYPERFINE_ITERATION} is not visible in prepare or conclude commands
I thought I could use the `${HYPERFINE_ITERATION}` env variable that is exposed (mentioned as a solution in https://github.com/sharkdp/hyperfine/issues/778) ; however this is only accessible in the main command of the iteration, not the prepare or conclude steps for each iteration.

So for example if I want to prepare , test and cleanup a series of files, I would do something like this (with `echo` for illustrative purposes) :
```
 ~/hyperfine --runs 5 --prepare 'echo touch file${HYPERFINE_ITERATION}.dat' \
                  'echo testing file${HYPERFINE_ITERATION}.dat' \
                  --conclude 'echo rm file${HYPERFINE_ITERATION}.dat' --show-output
```

However this does not work because the env var is not exposed in all the steps:
```
Benchmark 1: echo testing file${HYPERFINE_ITERATION}.dat
touch file.dat
testing file0.dat
rm file.dat
touch file.dat
testing file1.dat
rm file.dat
touch file.dat
testing file2.dat
rm file.dat
touch file.dat
testing file3.dat
rm file.dat
touch file.dat
testing file4.dat
rm file.dat
  Time (mean ± σ):      90.7 µs ± 202.8 µs    [User: 116.6 µs, System: 252.7 µs]
  Range (min … max):     0.0 µs … 453.5 µs    5 runs
```

so it's not possible to prepare and clean up each file.

alexchenai:
## Summary

- Fixes #781: `$HYPERFINE_ITERATION` is now available in `--prepare` and `--conclude` commands, matching the value used by the corresponding benchmark command.
- Previously, prepare/conclude commands always ran with `BenchmarkIteration::NonBenchmarkRun`, which meant the env var was never set.
- `--setup` and `--cleanup` commands are unaffected (they run once, not per-iteration).

## Changes

- Derive `Clone, Copy` on `BenchmarkIteration` so the same iteration value can be passed to prepare, benchmark, and conclude.
- Thread the `BenchmarkIteration` parameter through `run_intermediate_command` → `run_preparation_command` / `run_conclusion_command`.
- Each call site now passes the matching iteration (e.g. `Warmup(i)` during warmup, `Benchmark(i)` during benchmarking).

## Test plan

- [ ] `hyperfine --runs 3 --prepare 'echo prep-$HYPERFINE_ITERATION' --conclude 'echo done-$HYPERFINE_ITERATION' --show-output 'echo run-$HYPERFINE_ITERATION'` should show matching iteration numbers across prepare, run, and conclude.
- [ ] `hyperfine --warmup 2 --runs 2 --prepare 'echo prep-$HYPERFINE_ITERATION' --show-output 'echo run-$HYPERFINE_ITERATION'` should show `warmup-0`, `warmup-1` during warmup and `0`, `1` during benchmark.
- [ ] Existing tests pass (`cargo test`).


wyf027:
Thanks for this PR! I verified the approach locally — `cargo test` passes.

CI may not have run on this branch yet. I opened #858 with an integration test + changelog entry on top of your implementation (happy to close #858 if you prefer to add the test here instead).

Suggested one-liner test case from the issue:
`hyperfine --runs 2 --warmup 0 --show-output --prepare 'echo prep-$HYPERFINE_ITERATION' --conclude 'echo done-$HYPERFINE_ITERATION' 'echo run-$HYPERFINE_ITERATION'`

wyf027:
Correction: follow-up PR with test is #874 (not #858).

PR #877, opened 2026-05-16
fix: include parameter columns in CSV when using --reference (fixes #852)

### LINKED ISSUE #852: CSV error when using parameters with --reference
This bug manifests when using `--parameter-scan` and `--reference` together, e.g.
```bash
hyperfine \
  --runs 1 \
  --reference "sleep 1" \
  --parameter-scan secs 2 3 \
  --export-csv out.csv 'sleep {secs}'
```

The first parameterised run after the reference gives the following error when complete:
```
Error: CSV error: found record with 9 fields, but the previous record has 8 fields
```

Looking at `out.csv`, I can see that the column name for the `secs` parameter is missing:
```csv
command,mean,stddev,median,user,system,min,max
sleep 1,1.00279931462,0,1.00279931462,0.0006699600000000003,0.0018847000000000004,1.00279931462,1.00279931462
```

I would expect the parameter columns to exist but their values for the reference run to all be missing, since there is doesn't seem to be a way to specify what their values are in the reference command.

wyf027:
## Summary

- Fixes CSV export when `--reference` is combined with `--parameter-scan` / `-L`.
- Derive parameter column names from the first result that has parameters (not `.first()`, which may be the reference row with an empty parameter map).
- Emit empty parameter fields for the reference row.

## Test plan

- [x] `cargo test --bin hyperfine test_csv test_csv_with_reference`
- [x] `cargo clippy -- -D warnings`
- [ ] CI

Closes #852


wyf027:
CI is green on all targets. Ready for review when convenient — thanks!

PR #899, opened 2026-06-01
docs: show how to benchmark across git branches

### LINKED ISSUE #606: Benchmark across different branches
When I'm working on performance, I tend to want to compare a baseline from our main branch against my current work. It'd be pretty handy to be able to specify branches at the CLI level to handle the switch. Maybe:

```
hyperfine --compare-branch main "rails runner true"
```

You can achieve something similar by including the git checkouts in the command:

```console
$ hyperfine "git switch main; DISABLE_SPRING=1 bin/rails runner true" "git switch performance-improvements; DISABLE_SPRING=1 bin/rails runner true"
Benchmark 1: git checkout main; DISABLE_SPRING=1 bin/rails runner true
  Time (mean ± σ):      8.690 s ±  0.876 s    [User: 4.835 s, System: 4.665 s]
  Range (min … max):    8.114 s … 11.106 s    10 runs

Benchmark 2: git switch performance-improvements; DISABLE_SPRING=1 bin/rails runner true
  Time (mean ± σ):      8.352 s ±  0.363 s    [User: 4.778 s, System: 4.621 s]
  Range (min … max):    7.950 s …  9.020 s    10 runs

Summary
  'git switch performance-improvements; DISABLE_SPRING=1 bin/rails runner true' ran
    1.04 ± 0.11 times faster than 'git checkout main; DISABLE_SPRING=1 bin/rails runner true'
```

xfocus3:
## Summary
- Add a README example for benchmarking the same command across Git branches.
- Use `--parameter-list` with `--setup 'git switch {branch}'` so branch switching is outside the timed command.
- Mention `git worktree` for benchmarking uncommitted changes.

Closes #606.

## Verification
- Confirmed in `src/benchmark/mod.rs` that `--setup` receives parameter substitution and runs before the set of timing runs.
- Ran `git diff --check`.


ded-furby (commented):
Nice example. One side effect that may be worth calling out explicitly in the surrounding text: after the benchmark finishes, the repo will still be left on the last branch from  because  mutates checkout state outside the timed command.\n\nSince the paragraph already warns about a clean working tree, an extra sentence about ending on the final branch would help avoid surprising users who run this in their main checkout instead of a disposable worktree.

ded-furby (commented):
Nice example. One side effect that may be worth calling out explicitly in the surrounding text: after the benchmark finishes, the repo will still be left on the last branch from `--parameter-list` because `--setup 'git switch {branch}'` mutates checkout state outside the timed command.

Since the paragraph already warns about a clean working tree, an extra sentence about ending on the final branch would help avoid surprising users who run this in their main checkout instead of a disposable worktree.

PR #902, opened 2026-06-19
Export empty CSV columns for reference command parameters

### LINKED ISSUE #852: CSV error when using parameters with --reference

lixitrixi:
Fixes #852

The problem was that the first command result was used to create the parameter CSV headers, even though this could be the reference command (which has no params). Here I find the first result with non-empty params and use it instead.

For the reference command, I export an empty value for each parameter to match the number of columns. IMO this is correct since the reference command has no parameter values. #877 uses lookup to print each column; I avoid this since iterating is cheaper and a reference command is very rare compared to parametrized commands.

### DIFF
--- a/src/export/csv.rs
+++ b/src/export/csv.rs
@@ -21,6 +21,7 @@ impl Exporter for CsvExporter {
     ) -> Result<Vec<u8>> {
         let mut writer = WriterBuilder::new().from_writer(vec![]);
 
+        let mut num_params = 0;
         {
             let mut headers: Vec<Cow<[u8]>> = [
                 // The list of times and exit codes cannot be exported to the CSV file - omit them.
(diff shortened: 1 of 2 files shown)


lixitrixi:
@sharkdp hiya, just bumping this as it's useful fix for my future benchmarking :)

PR #904, opened 2026-06-29
[codex] Truncate export files when rewriting

kmg0308:
## Summary
- Create export files with truncation so rewriting a shorter result does not leave stale trailing content.

## Validation
- export tests, markdown export smoke check, cargo fmt check, and diff check passed in Docker.

## Notes
- Opened as a draft PR for maintainer review.


vadcx:
OpenOptions has a truncate option: https://doc.rust-lang.org/std/fs/struct.OpenOptions.html#method.truncate

PR #908, opened 2026-07-16
Add line-backed parameter values with --parameter-file

### LINKED ISSUE #813: Allow getting parameters from file lines

220nightmore-spec:
## Summary

- add `--parameter-file <VAR> <FILE>` for using LF- or CRLF-delimited file lines as parameter values
- keep file-backed values and the generated command product lazy, retaining only the active values while commands are executed
- preserve existing `--parameter-list` Cartesian-product order and allow list/file sources to be interspersed
- reject duplicate parameter names across list and file sources and detect files that change after the initial count pass
- document the option in CLI help, README, and the man page

The initial file pass counts values so command-name and per-command option validation still know the benchmark count. Iteration then reopens the file and reads one line at a time. Benchmark results remain retained as before for comparison and export; this change removes eager retention of the input file values and generated `Command` product.

## Validation

- `cargo fmt -- --check`
- `cargo +1.88.0 clippy --locked --all-targets`
- 63 unit tests passed on the declared Rust 1.88 MSRV
- 20 execution-order tests passed, including mixed CRLF file/list product ordering
- 34 Windows-compatible integration tests passed; two existing Unix-shell-specific tests were filtered locally
- `cargo +1.88.0 check --locked --target i686-pc-windows-msvc`
- `cargo build --locked --release`
- rendered CLI help assertions for the new option and LF/CRLF behavior
- `git diff --check`

Fixes #813.


Scripter17:
Since Sharkdp, the owner of Hyperfine, seems to no longer be maintaining it, I've decided that as long as this PR does what I need it to do and neither of the other two get merged before August 18th, I'll pay out the full bounty to all 3 of the people who've submitted PRs for this.

Scripter17:
Finally getting around to testing this and it seems pretty good!

Let me know where to send the money (preferably paypal but I'm flexible) and I'll send it over on/by the 18th.

PR #910, opened 2026-07-16
Add streaming parameter files

### LINKED ISSUE #813: Allow getting parameters from file lines

LeonDong0807:
Closes #813

## Summary

- add `--parameter-file VAR FILE` for reading parameter values from file lines
- process values with buffered, line-by-line I/O instead of materializing the file contents
- support both LF and CRLF line endings
- preserve command naming, reference-command ordering, and per-command export behavior
- document the new option and add unit/integration coverage, including a large input case

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy --locked --all-targets` (only pre-existing warnings)
- `cargo test --locked` — 105 tests passed


LeonDong0807:
The current CI run is marked `action_required`, which appears to be waiting for approval because this is my first contribution to the repository. When convenient, could a maintainer approve the workflow run? The full local suite passes (105 tests), and I’ll address any CI or review feedback promptly.

Scripter17:
I should've checked this before setting the bounty, but the last 3 commits to Hyperfine were 3, 5, and 8 months ago. It seems to be largely abandoned and/or done. There's a decent chance Sharkdp will just never get to this.

So how about this: I'll check out this patch at some point soon, and if I find it good and Sharkdp doesn't respond by, idk August 18th, I'll just pay you the bounty anyway. Technically the agreement was payment upon merge but it would be just rude for me to not pay out such a small bounty over something you can't control.

LeonDong0807:
Thank you — I really appreciate that. I’ll keep the PR ready and respond promptly to any feedback when you review it.

Scripter17:
Finally getting around to testing this, and I found a couple issues:

1. You can only provide one `--parameter-file` total, whereas if you specify multiple `--parameter-list`s Hyperfine uses every combination of each list. `--parameter-list a a1,a2,a3 --parameter-list b b1,b2,b3` runs the command with a1 and b1, a1 and b2, etc.. This isn't a major issue but is a surprising limitation.

2. Passing in a large file loads the entire file into memory at once. That is a major issue.

Edit: The code also says --parameter-file is incompatible with --parameter-list? I get being incompatible with --parameter-scan because --parameter-list is for some reason, but being incompatible with --parameter-list is just strange.

PR #914, opened 2026-08-04
Neutralize spreadsheet formula characters in CSV export

### LINKED ISSUE #915: CSV export writes command and parameter values unescaped, allowing spreadsheet formula injection when the exported file is opened
## Summary

`hyperfine --export-csv` writes each benchmarked command's name and every
`--parameter-list`/`--parameter-scan` value into the CSV file with no
protection against spreadsheet formula interpretation
(`CsvExporter::serialize`, src/export/csv.rs). Excel, LibreOffice Calc,
and Google Sheets all treat a cell beginning with `=`, `+`, `-`, or `@` as
a formula to evaluate on open, independent of the file's `.csv`
extension. Since both the command string and any parameter value are
attacker-influenceable in workflows where hyperfine parameterizes a
benchmark over externally-submitted names (for example, comparing a list
of community-submitted package or command variants in CI), a benchmark
run driven that way and exported to CSV for someone else to review
produces a file that executes an attacker-chosen formula the moment that
person opens it. This is CWE-1236.

Validated against v1.20.0.

## POC

(available upon request)

## Impact

Confirmed: a `--parameter-list payload '=1+1'` run exported to CSV
contains the literal, unescaped text `=1+1`. Opening the file in
LibreOffice Calc (verified via headless conversion) evaluates the cell
and shows `2`, confirming the exported value is interpreted as a live
formula rather than displayed as text. A stronger payload is a documented
extension of the same mechanism: a `HYPERLINK`/`WEBSERVICE`-style formula
can exfiltrate other cell contents to an attacker-controlled URL, and
legacy DDE formula syntax has historically enabled command execution in
older Excel versions. Neither was tested here; confirming formula
evaluation was sufficient to establish the defect.

## Suggested Fix

Before writing a value into a CSV cell, check whether it begins with `=`,
`+`, `-`, `@`, a tab, or a carriage return, and if so, prefix it with a
single quote so spreadsheet applications display it as literal text
instead of evaluating it. This is the standard mitigation for CWE-1236.

carfeii:
Fixes #915.

## Summary

`CsvExporter::serialize` (src/export/csv.rs) writes the benchmarked
command name and every `--parameter-list`/`--parameter-scan` value
directly into CSV cells with no protection against spreadsheet formula
interpretation. Excel, LibreOffice Calc, and Google Sheets all treat a
cell beginning with `=`, `+`, `-`, or `@` as a formula to evaluate on
open, independent of the file's `.csv` extension. Since both the command
string and any parameter value can be attacker-influenced (for example, a
benchmark parameterized over externally-submitted names in a CI job), an
exported CSV given to someone else to review could execute an
attacker-chosen formula the moment they open it. This is CWE-1236.

## Fix

Adds `neutralize_formula`, which prefixes a value with a single quote if
it begins with a formula-triggering character (`=`, `+`, `-`, `@`, tab, or
carriage return), so spreadsheet applications display it as literal text
instead of evaluating it. Applied to both the command field and every
parameter value in `CsvExporter::serialize`. This is the standard
mitigation for CWE-1236.

## Testing

- `cargo test --release` passes in full across all suites (unit,
  `common`, `execution_order_tests`, `integration_tests`).
- Added `test_csv_formula_injection_is_neutralized`, asserting a command
  starting with `=` and a parameter value starting with `@` are both
  prefixed with a single quote in the exported CSV.
- Manually reproduced the injection against an unpatched build: ran
  `hyperfine --parameter-list payload '=1+1' --export-csv out.csv --
  'echo hi {payload}'`, confirmed the raw file contained the literal,
  unescaped text `=1+1`, then converted it with LibreOffice Calc headless
  (`soffice --headless --calc --convert-to ods`, then back to CSV) and
  confirmed the cell evaluated to `2`. Repeated the same steps against
  this branch: the raw file now contains `'=1+1`, and the value survives
  the LibreOffice round-trip unchanged (still `'=1+1`), confirming it is
  displayed as text rather than evaluated.


carfeii:
Confirmed PR creation works, closing this test PR.

